PR #16, opened 2023-06-13
Add gpt model model parameter

KiranPranay:
By defualt, the model would be gpt-4.
But, the user can specify, the model they want.

### DIFF
--- a/main.py
+++ b/main.py
@@ -16,8 +16,8 @@ app = typer.Typer()
 @app.command()
 def chat(
     project_path: str = typer.Argument(None, help="path"),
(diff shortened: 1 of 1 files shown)


abhishekASB:
wew

ilutoo:
not sure about this, model gpt-4 can not work on proff lvl

ilutoo:
sometimes is work ye, but its mistacke of galagramos inside the knek

ScorchChamp:
It already works if you run main.py --model gpt-3.5-turbo (so this PR is useless)

PR #28, opened 2023-06-14
Update ai.py

AayushTech:
Just adding the comments in the code for people to understand what exactly is going on

### DIFF
--- a/ai.py
+++ b/ai.py
@@ -1,28 +1,27 @@
 
 import openai
 
-
 class AI:
     def __init__(self, **kwargs):
         self.kwargs = kwargs
 
     def start(self, system, user):
         messages = [
-                {"role": "system", "content": system},
-                {"role": "user", "content": user},
-            ]
+            {"role": "system", "content": system},  # System message
+            {"role": "user", "content": user},      # User message
+        ]
 
         return self.next(messages)
 
     def fsystem(self, msg):
-        return {"role": "system", "content": msg}
+        return {"role": "system", "content": msg}  # Construct system message object
     
     def fuser(self, msg):
-        return {"role": "user", "content": msg}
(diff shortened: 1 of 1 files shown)


w3bdesign (approved):

offiub:
Okay it can’t approved now because of the errors and issues . Could you
like us to rectify it sir

ai.py
```diff
@@ -36,4 +35,11 @@ def next(self, messages: list[dict[str, str]], prompt=None):
...
-        return messages + [{"role": "assistant", "content": "".join(chat)}]
\ No newline at end of file
+        return messages + [{"role": "assistant", "content": "".join(chat)}]  # Add assistant's message to chat history
+
+# Create an instance of the AI class
+ai = AI()
+
+# Example usage:
+# Start a conversation with a system message and a user message
+ai.start("Hello, how can I assist you?", "Can you provide information about your products?")
```
w3bdesign, line 45: What's the point of these lines?

w3bdesign (changes requested):
Those lines should be removed

ai.py
patillacode: I believe @AayushTech meant those as an example?

They don't belong here though, maybe in the `README.md` but I would get rid of them all together.
Also, the comments throughout the code are not that helpful, I believe comments should only exist in particular cases where the code for some reason is difficult to understand.

In this case I believe the code is simple enough

AntonOsika:
I think it's self explanatory for senior developers, will close, but thanks for trying to make it easier for beginners @AayushTech

w3bdesign:
Protip: 

You can use chatGPT to explain code if you feel something is hard to understand. You can also ask it to explain it in a non-technical way, to simplify the language.

PR #29, opened 2023-06-14
Add .env support

yhyu13:
1, Use dotenv to init env variables
2, Support custom openai base URL 
3, Tested locally
4, Update readme

### DIFF
--- a/main.py
+++ b/main.py
@@ -8,7 +8,40 @@ from ai import AI
 from steps import STEPS
 from db import DB, DBs
 import typer
+from dotenv import load_dotenv
 
+load_dotenv()
+
+api_key = os.getenv("OPENAI_API_KEY") 
+# Check if OPENAI_API_KEY is defined
+if api_key == None:
+    # Raise an error if OPENAI_API_KEY is not defined
+    raise ValueError("OPENAI_API_KEY is not defined")
+else:
+    # Set the OPENAI_API_KEY to the api_key variable
+    openai.api_key = api_key
+
+# Get the OPENAI_API_BASE
+api_base = os.getenv("OPENAI_API_BASE")
+# Check if OPENAI_API_BASE is defined
+if api_base != None:   
+    # Set the OPENAI_API_BASE to the api_base variable
+    openai.api_base = api_base
+
+# Get the list of valid GPT models
+valid_gpt_modles = [
+    "gpt-4",
+    "gpt-3.5-turbo"
+]    
+# Get the GPT model from the environment variable
+gpt_model = os.getenv("gpt_model")
+# Check if the GPT model is valid
+if gpt_model not in valid_gpt_modles:
+    # Raise an error if the GPT model is not valid
+    raise ValueError(f"{gpt_model} is not a valid model. Valid models are: {valid_gpt_modles}")
+
+# Get the GPT temperature from the environment variable
+gpt_temperature = os.getenv("gpt_temperature")
 
 app = typer.Typer()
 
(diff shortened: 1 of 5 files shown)


yhyu13:
Used gpt3.5 default to save money (its token price further reduced recently)

main.py
```diff
@@ -8,7 +8,40 @@
...
+
+# Get the OPENAI_API_BASE
+api_base = os.getenv("OPENAI_API_BASE")
+# Check if OPENAI_API_BASE is defined
+if api_base != None:   
+    # Set the OPENAI_API_BASE to the api_base variable
+    openai.api_base = api_base
+
+# Get the list of valid GPT models
+valid_gpt_modles = [
```
patillacode, line 32: Nothing critical but there is a typo here! 
`valid_gpt_models` not `valid_gpt_modles`

main.py
```diff
@@ -8,7 +8,40 @@
...
+    "gpt-4",
+    "gpt-3.5-turbo"
+]    
+# Get the GPT model from the environment variable
+gpt_model = os.getenv("gpt_model")
+# Check if the GPT model is valid
+if gpt_model not in valid_gpt_modles:
+    # Raise an error if the GPT model is not valid
+    raise ValueError(f"{gpt_model} is not a valid model. Valid models are: {valid_gpt_modles}")
```
patillacode, line 41: same here

patillacode:
I like the fact that you added the temperature setting btw

main.py
w3bdesign: Support for gpt-3.5-turbo-16k should be added as it is the new 16k model

offiub:
Okay it can’t approved now because of the errors and issues . Could you
like us to rectify it sir

main.py
yhyu13: https://github.com/AntonOsika/gpt-engineer/pull/29/commits/b5ed9ecaf658848f7dd3a6a429e8eb63f2fd5850

Model validation is removed, probably better to just let users figure out which model they want to use

yhyu13:
https://github.com/AntonOsika/gpt-engineer/commit/b5ed9ecaf658848f7dd3a6a429e8eb63f2fd5850

Model validation is removed, probably better to just let users figure out which model they want to use

main.py
```diff
@@ -8,7 +8,30 @@
 from steps import STEPS
 from db import DB, DBs
 import typer
+from dotenv import load_dotenv
 
+load_dotenv()
+
+api_key = os.getenv("OPENAI_API_KEY")
```
spazm, line 15: We can use `os.environ[key]` rather than `os.getenv(key)` to look up mandatory keys.

`environ[key]` raises a KeyError if key does not exist in the environment.  `KeyError(key)`

The current code does allow setting a custom exception message.

patillacode: I don't see much difference @spazm 

I mean @yhyu13 is checking if the kye exists or not below, in line 17.

Although I agree something like this would be shorter and easier to read/follow thoughts?

```python
try:
    openai.api_key = api_key
except KeyError:
    raise ValueError("OPENAI_API_KEY is not defined")
    
```

@yhyu13 can you add something like this?

main.py
```diff
@@ -8,7 +8,30 @@
...
+api_base = os.getenv("OPENAI_API_BASE")
+# Check if OPENAI_API_BASE is defined
+if api_base != None:   
+    # Set the OPENAI_API_BASE to the api_base variable
+    openai.api_base = api_base
+
+# Get the GPT model from the environment variable
+gpt_model = os.getenv("gpt_model")
+# Get the GPT temperature from the environment variable
+gpt_temperature = os.getenv("gpt_temperature")
```
patillacode, line 34: For these we could fall back into a default value:

```python
openai.api_base = os.getenv("OPENAI_API_BASE", "https://api.openai.com/v1")
gpt_model = os.getenv("gpt_model", "gpt-3.5-turbo")
gpt_temperature = os.getenv("gpt_temperature", 0.1)
```

RemkoDelsman: line 16 in https://github.com/AntonOsika/gpt-engineer/blob/12cb93fc7a6ea3d1865b6dd8a2f892c4105bebaf/scripts/rerun_edited_message_logs.py also has hard-coded reference to GPT-4...

patillacode:
Open discussion about this in #135

PR #36, opened 2023-06-14
Added README.md generation and enabled common markdown in files

jacoballessio:
After creating all the files, the program now generates a README.md file with instructions on how to use the program. Changed file syntax from being wrapped in triple back ticks to a custom syntax("<#-[CONTENT]-#>") to allow files to be generated with common markdown.


patillacode:
Hello @jacoballessio 

You added lots of files that shouldn't be there, basically everything under these folders:
- `__pychache__`
- `snake_test`

Maybe you can update the `.gitignore` too!

PR #44, opened 2023-06-15
Add code pipeline and optional structure for packaging it for release

montao:
Hej Anton, thanks for a cool project. This fork and PR contain github action workflow and packaging of an artifact. They are not perfect, but there may be something in there for you. I wanted to package it as an app to make it faster to try and use, e.g., just "import gptengineer" at the python console or even as a standalone CLI tool, `$ gptengineer.` Then I will try it myself to see if it can make a text adventure game similar to Zork or something.

### DIFF
--- a/ai.py
+++ b/ai.py
@@ -5,6 +5,14 @@ class AI:
     def __init__(self, **kwargs):
         self.kwargs = kwargs
 
+        try:
+            openai.Model.retrieve("gpt-4")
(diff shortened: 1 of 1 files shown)


AntonOsika:
Nice!

I think we need to make it a proper CLI tool first:

Currently the `identity` files are kind of a part of the application...

Do you see a way?

montao:
Closing this to make it more cohesive later

PR #38, opened 2023-06-14
Fixes for #1

Jeadie:
Was hoping to add some functionality to gpt-engineer, but wanted #1  merged beforehand. 

These should up to date with main. We should squash and merge due to number of commits.

PR #52, opened 2023-06-15
Minor clean up

patillacode:
Hi @AntonOsika 

I decided to apply a minor clean up and add a few details for the near future.

- I have pinned the requirements for now until we decide on a better approach to handle dependencies (I suggest [pip-tools](https://pip-tools.readthedocs.io/en/latest/))

- I have added a `pyproject.toml` file that contains `ruff` and `black` configurations, I believe that having these can help us maintain a cleaner codebase. (notice I have ran `black` and `ruff` on the existing codebase, no functionality has been changed)

- I have updated the `.gitignore` file, I have kept it minimal for now.

For the near future I was thinking a few tasks I can do, of course these are just suggestions, let me know your thoughts:

- write a proper `Makefile` for easier installation/run.
- setup `pip-tools`
- setup `pre-commit` hooks to enforce the use of `ruff` and `black`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,33 @@
+# See https://help.github.com/ignore-files/ for more about ignoring files.
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# Distribution / packaging
+dist/
+build/
+*.egg-info/
+*.egg
+
+# Virtual environments
+venv/
+ENV/
+
+# IDE-specific files
+.vscode/
+
+# Compiled Python modules
+*.pyc
+*.pyo
+*.pyd
+
+# macOS specific files
+.DS_Store
+
+# Windows specific files
+Thumbs.db
+
+# this application's specific files
 archive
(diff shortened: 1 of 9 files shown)


AntonOsika:
Merging!

As for Makefile, what would it do for us today?

pre-commit: Yes! Sounds great.

pip-tools: I mentioned this in #Looks good!

I just would like to merge #38, what does it do for us today?

patillacode:
Makefiles are usually a nice tool to abstract the commands a bit, for example, a classic way of using it is to create an `install` command that would create the `venv` and install the dependencies in `requirements.txt` 
so a user can potentially clone the repo, run `make install` and good to go.

We could add another command for example `make start` or `make run`, etc.

At the moment maybe we don't have many things to do besides "run the app" but a makefile could come in handy in the near future.

I read your comment regarding `pip-tools` in #38 let's keep it simple for now, classic `requirements.txt` will work, we only have a few dependencies.

I'll create the issue for the `pre-commit` setup and assign to myself.

PR #60, opened 2023-06-15
Make gpt-engineer pip installable/runnable

Jeadie:
## Feature
Make it easy to run
```shell
pip install .  # eventually pip install gpt-engineer
gpt-engineer <PATH> # or gpt-engineer --help
```

## Changeset
- improve pyproject.toml with project configs
- make `gpt_engineer/` directory and move python files
- Fix paths in source code.


Jeadie:
Example output
```shell
(.venv) jeadie@Jacks-MBP gpt-engineer % gpt-engineer --help
Usage: gpt-engineer [OPTIONS] [PROJECT_PATH]

Arguments:
  [PROJECT_PATH]  path

Options:
  --run-prefix TEXT               run prefix, if you want to run multiple
                                  variants of the same project and later
                                  compare them
  --model TEXT                    [default: gpt-4]
  --temperature FLOAT             [default: 0.1]
  --install-completion [bash|zsh|fish|powershell|pwsh]
                                  Install completion for the specified shell.
  --show-completion [bash|zsh|fish|powershell|pwsh]
                                  Show completion for the specified shell, to
                                  copy it or customize the installation.
  --help                          Show this message and exit.
  ```

File: gpt_engineer/db.py [hunk 1]
```diff
@@ -17,10 +18,14 @@ def __setitem__(self, key, val):
         with open(self.path / key, 'w', encoding='utf-8') as f:
             f.write(val)
 
+    def __contains__(self, key):
+        return (self.path / key).exists()
+
```
patillacode, line 23: Good catch.

gpt_engineer/main.py
```diff
@@ -1,11 +1,12 @@
...
 import typer
 
-from ai import AI
-from db import DB, DBs
-from steps import STEPS
+from gpt_engineer.chat_to_files import to_files
+from gpt_engineer.ai import AI
+from gpt_engineer.steps import STEPS
+from gpt_engineer.db import DB, DBs
+
```
patillacode, line 9: I also like this better, explicit and probably saving someone from a circular import dependency.

gpt_engineer/main.py
```diff
@@ -34,16 +36,20 @@ def chat(
...
         dbs.logs[step.__name__] = json.dumps(messages)
 
 
-if __name__ == "__main__":
+def execute():
     app()
+
+
+if __name__ == "__main__":
+    execute()
```
patillacode, line 55: can't we just call `app()` instead of `execute()` and get rid of the `execute` definition?

patillacode (changes requested):
Hey @Jeadie 

this looks great, although I left a couple of comments, only one requesting a change though.

couple of notes:

- can you make sure you get rid of all those old commits? I think you branched out in a weird state maybe?

- can you add the corresponding documentation to the `README.md` file?

~I'll give this a go in the meantime.~

EDIT:
I tried to run it, unsuccessfully.
It seems like running via the command doesn't work.

The error is this one:
```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/dvitto/projects/gpt-engineer-jeadie-60/venv/lib/python3.11/site-packages/identity/qa'
```

Running with `--help` gave the expected output but 
I believe it could be related to what @AntonOsika mentioned [here](https://github.com/AntonOsika/gpt-engineer/pull/44#issuecomment-1593519703).

Also, I just noticed, the parameters are not being taken care of.
Meaning passing parameters are not taken into account at all, am I missing something?


Thanks a lot.

gpt_engineer/main.py
Jeadie: This is helpful for referencing it in pyproject.toml

Jeadie:
I think it will be easier (and cleaner/best practice going forward) to squash and merge all contributions to have a more readable main branch. Making some improvements in a second PR, will QA again today.

Jeadie:
Fixes are up. They were just due to how i was testing things initially. Having identity as a file is highly fragile. We will need to move off it soon.

patillacode (changes requested):
It is now working, thanks for the changes.

Two more things:
- can you squash your commits?
- parameters (i,e `--model`) are not being taken care of, I believe you left that on purpose, is this something you plan on adding?

I don't feel comfortable merging something that tells the user "use me like this" (when running `gpt-engineer --help`) but it is not actually working unless we plan on adding it shortly.

Jeadie:
I did not touch `--model` flag and  `--model` flag seems to work on mine (passing an invalid value to `--model invalid_id` throws the correct error). We can squash commits at merge time, which is generally an easy approach.

PR #75, opened 2023-06-16
Update ai.py

Mtemi:
update on next function:  List[Dict[str, str]]

### DIFF
--- a/gpt_engineer/ai.py
+++ b/gpt_engineer/ai.py
@@ -1,19 +1,21 @@
 import openai
+from typing import List, Dict, Union
 
 
 class AI:
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs: Union[str, int, bool]):
         self.kwargs = kwargs
 
         try:
+            # Check if gpt-4 model is available
             openai.Model.retrieve("gpt-4")
         except openai.error.InvalidRequestError:
-            print("Model gpt-4 not available for provided api key reverting "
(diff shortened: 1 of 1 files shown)


gpt_engineer/ai.py
```diff
@@ -37,8 +39,9 @@ def next(self, messages: list[dict[str, str]], prompt=None):
 
         chat = []
         for chunk in response:
-            delta = chunk["choices"][0]["delta"]
-            msg = delta.get("content", "")
+            delta = chunk['choices'][0]['delta']
+            msg = delta.get('content', '')
```
patillacode, line 43: please revert back to double quotes.

I have seen this change back and forth, it is based on personal preference now but when we add `pre-commit` you will be forced to use doubles since we are going to run `black`

patillacode (changes requested):
Hello @Mtemi 

Thanks for your contribution.

I am not sure what you wanted to do here, I am guessing just add some typings?

If we are going to go down this route we should add it all over the code, not just one file, WDYT?

gpt_engineer/ai.py
Mtemi: that's fine

patillacode:
Thank your @Mtemi 

If you plan on adding this across the board please keep the PR open, we would be happy to review and merge when it's done.
Otherwise, please close the PR.

mandervrik (commented):
I want to make a NFT

PR #80, opened 2023-06-16
Makefile

patillacode:
By using a `Makefile` we can abstract some of the things needed to run the app or install it and it is a great place for future commands (like testing)

See a demo below:
[asciicast](https://asciinema.org/a/vuTLDIxS1n6hsA9s3nHGqh9kh)


jebarpg:
So I also see a lot of projects us a bat file in addition to a makefile to do installations. What are your thoughts about creating a batch file to run and install?

patillacode:
Sounds good to me although I am not familiar with those, we can add that one in another PR.

I believe `.bat` files are for Windows, right?

patillacode:
@AntonOsika @FOLLGAD 

WDYT?

If you would like to have something like this let me know and I will amend the README for it

I would suggest having two way of installing, classic `python` way (like it is now in the docs) but also the `make` way which abstract a bit and makes things easier for no _pythonistas_.

AntonOsika:
Thanks @patillacode!

I looked yesterday and wanted to ask, do you want to put black commands and pytest commands in here?

Or do you have a simple command to run all the tools in pyproject.toml you prefer?

AntonOsika:
Also my only concern with Makefiles...

If we change how things are run, we have to keep Makefile in sync.

(And it makes it "harder to find things" since there are so many files)

patillacode:
Thanks @AntonOsika 

- Yes, sometimes it is a bit hard to maintain if we don't keep it simple.
I'd say for now just using the Makefile as an alternative to installing would be enough.

- As for running `black` for example, I leave that to the `pre-commit` hooks, I don't think we need a `make black` command to run `black`

- For running tests, yes, we should add a `tests` command so when you do `make tests` it runs them.

- I will edit the `README.md` so people know there is now a Makefile that they can use to install and run the project

shubham-attri:
@patillacode @AntonOsika  I can document the makefile because it is very crucial step in setting up the codebase for development and a proper documentation is needed

PR #83, opened 2023-06-16
allow dotenv in ai, add ai unit tests, add CONTRIBUTING.md

rwmartinez:
I got tired of entering my API key every time I opened a new terminal and wrote some tests


rwmartinez:
@patillacode @AntonOsika This is a great repo, thanks for the hard work maintaining it.

This MR adds the ability to load the `OPENAI_API_KEY` from `.env` instead of only via the CLI. It also separates app vs. dev dependencies and added contributing.md details. I saw you had `.github/contributing.md` in another PR so this should beef it up a bit.

patillacode:
Open discussion related to this in #135

AntonOsika:
Thanks @rwmartinez 

For now, I prefer not using monkeypatch.

Also I think we are fine to just use env variables to keep it as simple as possible for now

PR #86, opened 2023-06-16
Mobina.sediighi

09125395122:
می خوام این پروفایل اینستاگرام را هم کنم


jebarpg:
این دقیقا چه کاری انجام می دهد؟
What does this do exactly?

patillacode:
I am sorry @09125395122 but this doesn't make sense to me, closing for now.

Please try to explain what you are providing in your PR in a clear way, if possible in English so it is easier for us to evaluate it.

Thanks

Issue #35, opened 2023-06-14
.py files are not being created. I just get all_output.txt that I manually have to create from.


mindwellsolutions:
Hi, I absolutely love this script. This is the most accurate auto-GPT development script I have tried yet, it's so powerful!

In the demo video it shows the script creating each of the development files, in my case .py files within the workspace folder automatically. My build isn't doing this I just get an all_output.txt file with all .py files codes in one place and a single python file.

How do I ensure that GPT-Engineer automatically creates the .py files for me. Thanks

kawpls:
i have same issue

Dawkinspv:
regex = r"`(.*?)`.*?```(python)?(.*?)```"


files = []
    for match in matches:
        path = match.group(1)
        # Get the code
       # code = match.group(1).split("\n")[1:]
       # code = "\n".join(code)
        filename = match.group(1)
        code = match.group(3).strip()


        # Add the file to the list
        files.append((path, code))

jebarpg:
Heres my solution:
```python
import re
import os

save_dir = "results/"
f = open("example/workspace/all_output.txt", "r")
s = f.read()
pattern = re.compile(r'^\*\*(.*?\.py)\*\*\s+```python\s+.*?(^(?:.*\n)*?)^```\s', re.MULTILINE)
os.makedirs(save_dir, exist_ok=True)
for (file_name, file_text) in re.findall(pattern, s):
    write_file = open(save_dir + file_name, "w")
    write_file.write(file_text)
    write_file.close()
    print(file_text, '*', file_name)
```
but sometimes you have to change the regex because the output isn't always the same.
so if you see **game.py** use the code above
if you see game.py: then use this:
`pattern = re.compile(r'^(.*?\.py):\s+```python\s+.*?(^(?:.*\n)*?)^```\s', re.MULTILINE)`

DrMemoryFish:
> Heres my solution:
> 
> ```python
> import re
> import os
> 
> save_dir = "results/"
> f = open("example/workspace/all_output.txt", "r")
> s = f.read()
> #pattern = re.compile(r'^\*\*(.*?\.py)\*\*\s+```python\s+.*?(^(?:.*\n)*?)^```\s*', re.MULTILINE) #Example **game.py**
> #pattern = re.compile(r'^(.*?\.py):\s+```python\s+.*?(^(?:.*\n)*?)^```\s*', re.MULTILINE) #Example game.py:
[...]

PR #87, opened 2023-06-16
Fixing chat parsing issues that result in wrong files

goncalomoita:
Improved parsing function (parse_chat) which directly affects the file generation process.

"parse_chat", which is located in [gpt_engineer/chat_to_files.py](https://github.com/AntonOsika/gpt-engineer/blob/main/gpt_engineer/chat_to_files.py), is now able to discern filenames by searching in specific areas near the code block.
Rather than performing filename pattern matching on the entire prompt completion output, the new approach looks for
the filename in logical spots (just like a human would).

Also included a tests file tests/test_chat_parser.py, that checks if parse_chat is working properly for a plethora of code formatting techniques.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,9 @@ Contributors welcome! If you are unsure what to add, check out the ideas listed 
 ## Features
 You can specify the "identity" of the AI agent by editing the files in the `identity` folder.
 
-Editing the identity, and evolving the main_prompt, is currently how you make the agent remember things between projects.
+Editing the identity, and evolving the `main_prompt`, is currently how you make the agent remember things between projects.
 
-Each step in steps.py will have its communication history with GPT4 stored in the logs folder, and can be rerun with scripts/rerun_edited_message_logs.py.
+Each step in `steps.py` will have its communication history with GPT4 stored in the logs folder, and can be rerun with `scripts/rerun_edited_message_logs.py`.
 
 ## Contributing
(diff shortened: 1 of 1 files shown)


NoCLin:
could you please take a look and test it? https://github.com/AntonOsika/gpt-engineer/pull/94

gm0l74:
I like this idea for self-reviewing and improving the code itself. I've talked about using this for parsing in [#35](https://github.com/AntonOsika/gpt-engineer/issues/35#issuecomment-1594508297):

"""
I thought about the problem of ensuring a consistent format when deciding whether or not to improve this parser. My initial thought was "Does better parsing make sense? Can I fix this with a "self-review" call to a LLM? Should I call the LLM just to find what the filenames are?".
Those are possible solutions but they add performance overhead, since you have to wait for another inference, and ultimately cost.
"""

But I encourage you to do this for a "self-reviewer", a sort-of senior programmer that fixes the mistakes of a junior 😆

goncalomoita:
I like this idea for self-reviewing and improving the code itself. I've talked about using this for parsing in [#35](https://github.com/AntonOsika/gpt-engineer/issues/35#issuecomment-1594508297):

"""
I thought about the problem of ensuring a consistent format when deciding whether or not to improve this parser. My initial thought was "Does better parsing make sense? Can I fix this with a "self-review" call to a LLM? Should I call the LLM just to find what the filenames are?".
Those are possible solutions but they add performance overhead, since you have to wait for another inference, and ultimately cost.
"""

But I encourage you to do this for a "self-reviewer", a sort-of senior programmer that fixes the mistakes of a junior 😆

PR #105, opened 2023-06-17
Fix an exception for GPT response with lang=sh

jjzhuo:


patillacode:
Nice call, thanks @jjzhuo

PR #22, opened 2023-06-14
Allow .env file variables to be used

ScorchChamp:
This fix allows the .env to be used for environment variables. 
(Especially useful for windows devices)

PR #106, opened 2023-06-17
Adding env support

Glifaus:
Added .env support to save OPENAI_API_KEY, and use it on ai.py class. Added simple test and updated requirements. Tested ok


patillacode:
Hi @Glifaus 

There have been several PR attempts for this functionality like #83 #29 or #22 

I like that you added tests, thanks, although I think your `requirements.txt` file is polluted, no need to add all those dependencies, I am guessing you only really added `python-dotenv` as requirement, right?

Anyway, I believe there will be some conversations with @AntonOsika to try to bring this feature in (might be this PR or maybe any of the others)

Glifaus:
Hi! You're right. Sorry I polluted the requirements. I used the freeze command to take all and leave is not the best option (I'm fairly new to python). I can upload a new requirements clean, but is what you comment.

patillacode:
Open discussion related to this in #135

PR #107, opened 2023-06-17
Update ai.py to use typing module

puniith09:
Updated ai.py to utilise the typing module for list and dict

### DIFF
--- a/gpt_engineer/ai.py
+++ b/gpt_engineer/ai.py
@@ -3,18 +3,18 @@ import openai
 
 
 class AI:
-    def __init__(self, **kwargs):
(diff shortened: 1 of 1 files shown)


patillacode:
Hi @puniith09 

We appreciate you PR but I think we should either go all-in with typing in every file, in every method or just avoid it.
Just changing this one line is not very valuable, WDYT?

If you want to fix all files with proper typing please feel free to add that to your PR.

puniith09:
That's great to hear, i would like to do that for whole project!

puniith09:
Hey @patillacode i think this is the only occurence left to use typing module in whole.

PR #112, opened 2023-06-17
Choose directory before execution

Snoup97:
adjust the `execute_workspace` function to prepend the directory before the command. #103


AntonOsika:
Thanks @Snoup97 

I think it doesn't fully work since `&&` will not work if it starts with a comment...

Since I already had a diff underway, I merged #123 this time!

Snoup97:
Hi @AntonOsika 
Thanks for the feedback.

PR #118, opened 2023-06-17
add workflow to execute unit tests

Snoup97:


.github/workflows/pytest-action.yml
```diff
@@ -0,0 +1,26 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.9
```
AntonOsika, line 19: We want at least 3.10

AntonOsika:
Can you explain what else is needed to make these tests run in the PR?

Snoup97:
@AntonOsika to enable workflows to run on PRs you need to head to the settings of this repo, to the actions/general section. there should be a section that says "Fork pull request workflows", here you just set the checkbox for the "Run workflows from fork pull requests". the other checkboxes can be left unchecked.

.github/workflows/pytest-action.yml
Snoup97: alright ill change it

Snoup97:
@AntonOsika ill also change it to run on pytest and move the tests i wrote in #128 to pytest.

patillacode:
If we are going to use pytest, we should add it as a requirement

patillacode:
Moving this into draft until `pytest` is added as a requirement, please mark as ready when done and we can approve and merge!

.github/workflows/pytest-action.yml
```diff
@@ -0,0 +1,26 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.10
```
patillacode, line 19: I think it need to be something like 3.10.11

build is throwing the error:
```
  Error: The version '3.1' with architecture 'x64' was not found for Ubuntu 22.04.
  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

Snoup97: yep. fixed it

carlthome:
There already is a CI workflow on main for automated unit testing in PRs with pytest.

PR #120, opened 2023-06-17
Improved chat parsing with no AI logic

goncalomoita:
This fixes compatibility issues in PR [#87](https://github.com/AntonOsika/gpt-engineer/pull/87) that have arisen from recent changes to the repository.
Additionally, an improvement to the parser unit tests has been done.

The goal of the this parser is to make sense of what the LLM has provided without requiring another inference call which adds time and cost overhead. My remarks on this matter were stated in [#35](https://github.com/AntonOsika/gpt-engineer/issues/35#issuecomment-1594508297).
Further debate has been conducted in PR [#94](https://github.com/AntonOsika/gpt-engineer/pull/94)

### DIFF
--- a/gpt_engineer/chat_to_files.py
+++ b/gpt_engineer/chat_to_files.py
@@ -1,22 +1,85 @@
 import re
+from typing import List, Tuple
+
+# Amount of lines within the code block to consider for filename discovery
+N_CODELINES_FOR_FILENAME_TA = 5
+
+# Default path to use if no filename is found
+DEFAULT_PATH = 'unknown.txt'
 
 
-def parse_chat(chat):  # -> List[Tuple[str, str]]:
-    # Get all ``` blocks
-    regex = r"```(.*?)```"
+def parse_chat(chat: str, verbose: bool = False) -> List[Tuple[str, str]]:
+    '''
+    Parses a chat message and returns a list of tuples containing
+    the file path and the code content for each file.
+    '''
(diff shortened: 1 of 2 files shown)


bmeddeb (approved):

patillacode:
Hi @goncalomoita 

thanks a lot for this PR.

I would like to merge this but I need confirmation/feedback from @AntonOsika and anyone else who wants to chime in.

I am thinking that we can merge this and remove it later if [#94](https://github.com/AntonOsika/gpt-engineer/pull/94#issuecomment-1595779838) gets done. I think #94 fixes this with functions though

AntonOsika:
- Solves a clear problem ✅
- Problem is significant ✅
- Solution is robust ✅

Looks good to me!

AntonOsika:
The only thing is that I always want simpler solutions ❌

Less code -> easier to change the repository later.

What we could do:

Use a system prompt as a final message that explicitly asks for the format we prefer.

EnzoMartin:
I had a variety of issues with chat parsing, I made similar improvements including supporting directory creation in #140

patillacode:
@AntonOsika 

Merging now since you agree, we can always simplify if possible later.

Just to be clear I think #94 is the way to go, it is simpler imo and uses the model itself via functions to solve this issue.

Until that is merged (if ever) this provides a solution for an issue that many users have at the moment.

patillacode:
Thank you @goncalomoita

PR #123, opened 2023-06-17
improve execution step

### LINKED ISSUE #103: Make the `execute_workspace` step more reliable by having the prompt instruct it to navigate to the right path first

AntonOsika:
Fix #103

### DIFF
--- a/gpt_engineer/steps.py
+++ b/gpt_engineer/steps.py
@@ -56,7 +56,7 @@ def gen_spec(ai: AI, dbs: DBs):
     '''
     Generate a spec from the main prompt + clarifications and save the results to the workspace
     '''
-    messages = [ai.fsystem(setup_sys_prompt(dbs)), ai.fsystem(f"Main prompt: {dbs.input['main_prompt']}")]
+    messages = [ai.fsystem(setup_sys_prompt(dbs)), ai.fsystem(f"Instructions: {dbs.input['main_prompt']}")]
 
     messages = ai.next(messages, dbs.identity['spec'])
     messages = ai.next(messages, dbs.identity['respec'])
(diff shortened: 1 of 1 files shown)


gpt_engineer/steps.py
```diff
@@ -119,7 +122,9 @@ def execute_workspace(ai: AI, dbs: DBs):
         return messages
     print('Executing the code...')
     print()
-    subprocess.run(command, shell=True)
+
+    # Run the subprocess in dbs.workspace.path
+    subprocess.run('bash run.sh', shell=True, cwd=dbs.workspace.path)
```
patillacode, line 127: Hey @AntonOsika 

I think I am missing something here, are we forcing the creation of a `run.sh` file somewhere?

jebarpg: might need an environment flag to choose which shell to use for this. something like 

```python
import subprocess
import psutil
import os
#https://github.com/giampaolo/psutil
p = psutil.Process(os.getppid())
env = p.environ()
print(env)
if 'COMSPEC' in env:
    subprocess.run(f'cmd run.bat', shell=True, cwd=dbs.workspace.path)
if 'SHELL' in env:
    subprocess.run(f'{env['SHELL']} run.sh', shell=True, cwd=dbs.workspace.path) #{env['SHELL']}will return the shell like this '/bin/bash'

```

patillacode:
With your permission @AntonOsika 
I am removing merged branches to avoid cluttering.

AntonOsika:
Great. I'm used to that happening automatically 🙏 @patillacode

gpt_engineer/steps.py
Rockin30: Can I change this in my own build? Super unfamiliar with github and the like, but I'm trying to run gp-engineer through vscode terminal and conda env.

PR #128, opened 2023-06-17
Extend tests for DB class

Snoup97:


gpt_engineer/db.py
```diff
@@ -11,14 +11,17 @@ def __init__(self, path):
...
+                return f.read()
+        except FileNotFoundError:
+            raise KeyError(key)
 
     def __setitem__(self, key, val):
         Path(self.path / key).absolute().parent.mkdir(parents=True, exist_ok=True)
 
         with open(self.path / key, 'w', encoding='utf-8') as f:
-            f.write(val)
+            f.write(str(val))
```
AntonOsika, line 24:
```suggestion
            f.write(val)
```

I think we should only allow str as input here, it's probably something wrong if it's not a string

tests/test_parse_chat.py
```diff
@@ -0,0 +1,22 @@
+import unittest
+from gpt_engineer.chat_to_files import parse_chat
+
+
+class TestParseChat(unittest.TestCase):
+
+    def test_parse_chat_with_valid_messages(self):
+        chat = "```path1.py\nline1\nline2```" \
```
AntonOsika, line 8: There should be newline between blocks, this looks weird!

AntonOsika:
Thanks @Snoup97.

I prefer pytest and was using it thus far.

Do you think we could move to it?

If you have a good reason, I'd be happy to merge like this too (and maybe change later).

Please see my comment on not allowing arbitrary types, only str, when writing!

AntonOsika:
Run ci and it should be good

PR #136, opened 2023-06-18
Add `python-dotenv`

NeuralEmpowerment:
1. Provide `.env` support to manage API keys.
2. Slight update to `identity/generate` which allows for the correct GPT filename output for non-python files. (i.e. before js files were being named `javascript`)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+OPENAI_API_KEY=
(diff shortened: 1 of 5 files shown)


patillacode:
Open discussion about this in #135

patillacode:
Will address if the discussion decided so.
Closing for now, thanks @NeuralEmpowerment

EnzoMartin:
IMO this is a really good addition

PR #140, opened 2023-06-18
Directory creation & project directory

EnzoMartin:
This does a few changes:

- [x] Add support for directory creation by having the path in the filename
- [x] Enhance the `DB` class to be able save binary files for adding an image generation step later
- [x] Moves the `main_prompt` files to be in a `projects` folder that is `.gitignore`'d for simplicity in using this repository directly to generate projects
- [x] Adds some wording to the philosophy to _attempt_ to generate all needed project files, though this seems to be hit and miss still
- [x] Adds an optional cleanup function to the script startup to delete the memory/logs/etc. before running the prompt

I've tested with the following languages using a Windows PC:

- Terraform
- JavaScript (NodeJS & ReactJS)
- Terraform
- Ansible
- Python3

Though I've been struggling to get it to consistently output a correct `package.json` and `requirements.txt` when it uses dependencies.

**NOTE:** I used ChatGPT to rewrite portions of this, especially with dealing with the RegEx as I can't RegEx to save my life


EnzoMartin:
I've tested with the following languages using a Windows PC:

- Terraform
- JavaScript (NodeJS & ReactJS)
- Terraform
- Ansible
- Python3


Though I've been struggling to get it to consistently output a correct `package.json` and `requirements.txt` when it uses dependencies

jebarpg:
> This does a few changes:
> 
> * [x]  Add support for directory creation by having the path in the filename
> * [x]  Enhance the `DB` class to be able save binary files for adding an image generation step later
> * [x]  Moves the `main_prompt` files to be in a `projects` folder that is `.gitignore`'d for simplicity in using this repository directly to generate projects
> * [x]  Adds some wording to the philosophy to _attempt_ to generate all needed project files, though this seems to be hit and miss still
> * [x]  Adds an optional cleanup function to the script startup to delete the memory/logs/etc. before running the prompt
> 
> I've tested with the following languages using a Windows PC:
> 
> * Terraform
> * JavaScript (NodeJS & ReactJS)
> * Terraform
> * Ansible
> * Python3
> 
> Though I've been struggling to get it to consistently output a correct `package.json` and `requirements.txt` when it uses dependencies.
> 
> **NOTE:** I used ChatGPT to rewrite portions of this, especially with dealing with the RegEx as I can't RegEx to save my life

I can help you with RegEx in the future, ping me when you need help.

AntonOsika:
Thanks @EnzoMartin for the PR!

To be honest, I think it is safer and cleaner to separate explicit delete step :)

Would love if you change as a part of a separate future PR?!

Regardless, I'll merge now to reduce conflicts with other upcoming merges 🎉

EnzoMartin:
Yeah good point, I expect the delete step is something that isn't needed at all later on tbh

The reason I added it is I was having some issues testing significant prompt changes 😁

PR #144, opened 2023-06-18
Model fallback

jebarpg:
Create a parent script model_select.py in the main repo directory to handle falling back to the gpt-3.5-turbo model for both the main.py and the rerun_edited_message_logs.py scripts. This centralizes where the model is handled and takes it out of the ai.py file so that both main.py and rerun_edited_message_logs.py can access and be updated by model_select.py. solving the issue people have who do not have access to gpt-4 with their current API key. In the future this makes it simple to edit in one place instead of having to manually edit two files with the same information.


w3bdesign:
@jebarpg 

Where is the array of fallback models?

jebarpg:
> @jebarpg
> 
> Where is the array of fallback models?

 I didn't place them in the files. I only added the gpt-3.5-turbo, but you can get a list of all fallback models from this openai API call:
 
```python
model_lst = openai.Model.list()

for i in model_lst['data']:
    print(i['id'])
```
 which will list all avaible models, but you have to suss out the chat models from the list. a better way is to go to your open ai account and check [rate limits page](https://platform.openai.com/account/rate-limits) and see a list of your available models for chat.. I just sent a request to OpenAI to add the ability to get a list of individual categories of models from their API endpoints (chat models, text models, image models, audio models, other models, etc...). When they create that then we can make a selection prompt for the user to choose from their models they have access to with their given API Key. 

here are the fallback models for those who do not have access to gpt-4: 
gpt-3.5-turbo
gpt-3.5-turbo-0301
gpt-3.5-turbo-0613
gpt-3.5-turbo-16k
gpt-3.5-turbo-16k-0613

PR #173, opened 2023-06-18
Fix square bracket file name issue

### LINKED ISSUE #169: Issues with writing file names
Hi, great tool!

I just starting trying it, and I'm running into some issues. 

Initially, I was encountering the same problem as detailed in this issue: https://github.com/AntonOsika/gpt-engineer/issues/167. Fortunately, I noticed you fixed this and after pulling the changes, that issue was resolved.

However, I'm now facing another issue related to file name generation. The program is formatting file names as '`filename`':
image

At other (most) times, it's formatting them as '[filename]', which is resulting in file outputs like this:

image

I suspect this might have something to do with this commit: https://github.com/AntonOsika/gpt-engineer/commit/8180f0346c8a380637b72212432aa9abd5c04a9c#diff-945067bb601867dd64fd5ae92de9bf829da8545700835ac49744902ddec355fd

I'm using gpt-3.5-turbo since I don't have access to gpt-4 yet. Maybe the model is not as smart to understand this prompt?

davejcameron:
Fixes an issue where file names surrounded by `[]` would not be stripped out properly in creating the file name. Also added some tests.

Handles the following cases
- [code.py] => code.py
- [id].jsx => [id].jsx

Fixes #169

### DIFF
--- a/gpt_engineer/chat_to_files.py
+++ b/gpt_engineer/chat_to_files.py
@@ -11,6 +11,9 @@ def parse_chat(chat):  # -> List[Tuple[str, str]]:
         # Strip the filename of any non-allowed characters and convert / to \
         path = re.sub(r'[<>"|?*]', "", match.group(1))
 
+        # Remove leading and trailing brackets
+        path = re.sub(r"^\[(.*)\]$", r"\1", path)
+
         # Get the code
         code = match.group(2)
 
--- a/tests/test_chat_to_files.py
+++ b/tests/test_chat_to_files.py
@@ -1,0 +1,94 @@
+import textwrap
+
+from gpt_engineer.chat_to_files import to_files
+
+
+def test_to_files():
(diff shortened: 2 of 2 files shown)


jebarpg:
> Fixes an issue where file names surrounded by `[]` would not be stripped out properly in creating the file name. Also added some tests.
> 
> _Note: This will likely be an issue for some files where brackets are legal like [id].jsx style file names._
> 
> Fixes #169

may there can be some logic for particular file extensions which allow this type of file name convention, other wise apply your fix.

davejcameron:
@jebarpg added some tests for the `[id].jsx` case and switched to a regex that checks for wrappings brackets.

AntonOsika:
Nice with the tests!

patillacode:
> These unstability issues that it works "most" of the time are still bugging me!
functions should solve this instability  (#94 )

PR #189, opened 2023-06-19
fix(bug): toml misconfiguration

gblikas:
"pre-commit: command not found" was generated upon clean codepsace; "pre-commit" was in /requirements.txt, but missing in .toml file.

**Steps to reproduce**
- create default Codespace 
- run `make install`
- See error: `pre-commit: command not found`


patillacode:
Thanks @gblikas

PR #197, opened 2023-06-19
add pathlib

subnwa:
The purpose of this code is to provide a simple class hierarchy for handling file system paths and common user operations on them.


subnwa:
I need to fix some things

PR #207, opened 2023-06-19
Handle python versions in Makefile

bannsec:
In many nix environments, `python` no longer exists as a symlink, and according to [PEP394](https://peps.python.org/pep-0394/) it's ambiguous as to what version it would be. Similarly, it's recommended to use `python -m pip` to ensure the correct `pip` is being used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ SHELL := /bin/bash
 COLOR_RESET=\033[0m
 COLOR_CYAN=\033[1;36m
 COLOR_GREEN=\033[1;32m
+PYTHON := $(shell command -v python3 2> /dev/null || echo "python")
 
 .PHONY: help install dev-install run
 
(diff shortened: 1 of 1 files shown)


patillacode:
Hello @bannsec 

I am conflicted.

I see your point but I also think that if you are going to use this application we would assume you do have python installed and in your path (symlink or not)

I'd say we tend to follow PEP recommendations but this one seems a bit odd.
(If someone is still using Python 2 btw, that's on them 😅 )

Also, `PYTHON := $(shell command -v python3 2> /dev/null || echo "python")` won't work on Windows AFAIK.

Regarding using `python3 -m`, why not `python -m`?

bannsec:
So for the python link, for instance, I tried setting this up with make
install on by jammy Ubuntu and it wouldn't run due no python symlink. I had
to manually edit the file to change to python3, which was the impetus of
automating that check.

Agree this wouldn't discover correctly on windows. Assuming this project
wasn't meant to run on windows due to the bash sub processing in steps.py.
I'm sure it can be updated to work on both platforms if that's desired.

The pip thing basically goes back to the same ambiguity of python. It's
possible there won't be a symlink for python (like my jammy), and similarly
it could be pointing to python2 if it exists (i.e. either older versions of
python or Ubuntu "update-alternatives", etc).

Or could just tell people to make a symlink? Though I generally prefer
automating discovery when possible, personally.

patillacode:
I agree but automations need to cover all cases, and this really doesn't.

I think we should leave as is and:
- assume the user has python or figures out he/she needs to install it
- assume python version > 2

bannsec:
For clarity, it has nothing to do with having python not installed. Many
distributions no longer have the command python linked to anything. This is
intentional to avoid confusion. Certainly your call on how to handle it,
just want to ensure you're not under the assumption that not having the
"python" command means python is not installed, as that's incorrect.

carlthome:
On Ubuntu, it's nice to `apt install python-is-python3` to have `python` available.

PR #208, opened 2023-06-19
Improved CONTRIBUTING.md

Harshad112:
corrected Typo Mistakes

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -98,7 +98,7 @@ fix end of files.........................................................Passed
 
 Now your file has been committed and you can push your changes.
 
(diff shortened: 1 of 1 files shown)


Harshad112:
General 1 suggestion here , can you please add CONTRIBUTING.MD file in home/front page , Because it is easier to find to NEW users or developers

patillacode:
thanks @Harshad112 

There is a link to the `CONTRIBUTING.md` file in the README (or it should be, I'll add it if not) 

From github documentation:
```
If a repository contains more than one CONTRIBUTING file, then the file shown in links is chosen from locations in the following order: the .github directory, then the repository's root directory, and finally the docs directory.
```

That is why it is there, leaving it there for now, thanks for the suggestion though.

PR #210, opened 2023-06-19
Add pytest pre commit hook

### LINKED ISSUE #205: Run pytest in pre-commit
- Add requirement to pyproject.toml
- Setup `.pre-commit-config.yaml` config
- test that everything is working with `pre-commit run` and in github actions

patillacode:
Closes #205 

- Added a pre commit hook to run `pytest` on commit
- Added requirement `termcolor` (used in 'print_chat.py')


patillacode:
I am not sure if I understood exactly what we need in the issue #205 , but we are now running the tests both in the `pre-commit` GH action and in the `pytest` GH action.

I would remove running tests on each commit and just leave it in the GH actions

I'll wait for feedback

patillacode:
I am no expert on GH actions.

I have invested more time than I would like to recognize trying to fix the `pre-commit` action in the CI.

As I mentioned before I wouldn't add the tests in as a hook, now I completely subscribe that because if we avoid running the tests as s pre-commit hook my problem with the CI goes away 😅 

I'm going to remove tests being run as a hook in favor of the dedicated action already running the tests.

Let me know if this work for you.

carlthome:
Generally speaking it's considered bad practice to run unit tests in pre-commit.

patillacode:
I agree @carlthome 

Leaving that to the CI. Merging, thanks!

PR #214, opened 2023-06-19
error fix #6a5ba65

subnwa:
Although the OPENAPI key was previously registered, it was getting an error and I could not find the solution. Now, with a small change, the problem has completely disappeared and now it works flawlessly.

### DIFF
--- a/gpt_engineer/main.py
+++ b/gpt_engineer/main.py
@@ -1,8 +1,9 @@
 import json
 import logging
 import os
-import pathlib
 import shutil
+
+from pathlib import Path
 
 import typer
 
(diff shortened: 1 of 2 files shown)


gpt_engineer/main.py
```diff
@@ -51,13 +51,12 @@ def chat(
...
     )
 
     for step in STEPS[steps_config]:
         messages = step(ai, dbs)
         dbs.logs[step.__name__] = json.dumps(messages)
 
-
 if __name__ == "__main__":
-    app()
+    app.run()
```
huntersinclairtiv, line 62: I get an error on this -  says typer run() does not exist

subnwa: Yeah I guess that's why it's giving an error. I need to find another solution for OPENAPI error

patillacode: Yeah, I don't think this works for me either.

Also, I think your branch is not up to date with `main` @subnwa

subnwa: I realized that there is an error in my own system, I will review it again.

AntonOsika:
I don't understand.

Feel free to open with a clearer description of what you are solving and how it solves it

PR #221, opened 2023-06-19
Additional filename handling

### LINKED ISSUE #211: File name is not correct
The generated py files is not named correct. 
It got the extra symbol `]` in the file extension name. 
image

davejcameron:
Ran through the benchmark script and discovered two cases from gpt-3.5 still wasn't handling file names properly. Added test cases and regex to fix these issues. 

Fixes #211


AntonOsika:
Nice!

PR #226, opened 2023-06-19
Makefile: Update to list files in projects/ in help.

servel333:
- Adds some more messages to `make help` for folders in projects/.
- Adds "make ..." to make help messages more like examples for copy-paste.
- Adds "##" for help messages to prevent full-line copy-paste issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,16 +14,16 @@ COLOR_GREEN=\033[1;32m
 name := $(word 2,$(MAKECMDGOALS))
 
 help::
-	@echo "Please use 'make <target>' where <target> is one of the following:"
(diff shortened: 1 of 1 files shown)


patillacode:
Hi @servel333, thanks for your interest!

I am sorry to say I am not sure this adds any value.

The last help line should be pretty self explanatory with `run <folder_name>`

I believe there is no need to list all potential `make run` commands available depending on your folders inside `projects/`

servel333:
That's fair.

servel333:
I just added a few lines that imports another makefile if it exists, and it works for me locally. Thanks for considering.

PR #244, opened 2023-06-20
Instruction for fish shell

w3n-grogu:
Solves #242

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ For **development**:
 - `cd gpt-engineer`
 - `make install`
 - `source venv/bin/activate`
+### For fish shell, run
+- `source venv/bin/activate.fish`
 
 **Setup**
 


patillacode:
Hi @w3n-grogu 

AFAIU @AntonOsika is a big fan of keeping the README to a minimum.
Based on that, if we add this we should add it for all terminals out there (quite a few) and it would end up polluting the README a bit.

I think if you are using `fish` you should probably "know about" this sort of stuff  since you probably find yourself in similar situations on the daily basis 

I appreciate your interest but I don't think this belongs in the README.

w3n-grogu:
> Hi @w3n-grogu
> 
> AFAIU @AntonOsika is a big fan of keeping the README to a minimum. Based on that, if we add this we should add it for all terminals out there (quite a few) and it would end up polluting the README a bit.
> 
> I think if you are using `fish` you should probably "know about" this sort of stuff since you probably find yourself in similar situations on the daily basis
> 
> I appreciate your interest but I don't think this belongs in the README.

agreed, makes sense.

PR #250, opened 2023-06-20
Handle newlines between filename

davejcameron:
I ran through the benchmark tests after the switch to simple, and it used `##` headings more commonly for me on generating the file contents. 

>     This is a sample program.
> 
>     ## file1.py
> 
>     ```python
>     print("Hello, World!")
>     ```

Adds two tests and handles the newline


davejcameron:
Side note, it would be nice if the results generated the tree structure of the workspace directory to make it easy to determine where odd files are created, or if files are missing.

patillacode:
Thanks @davejcameron 

PS: @AntonOsika I am merging this, yet another example of #94 coming in handy.

PR #258, opened 2023-06-20
Added Description

Gabekess:


jontonsoup4:
`requirements.txt` is understood as the standard for Python requirements. There's no need to add an additional comment to this file

PR #278, opened 2023-06-21
Create calculator_prompt Benchmark

LopeKinz:
A uniqe Benchmark to test chatgpt's abilities in UI Design and Tkinter usage


patillacode:
If you want to add such benchmark I would suggest you do the whole thing, like the other benchmarks.

If you wish to reopen the issue please do following the new issue template.

thanks @LopeKinz

PR #292, opened 2023-06-21
fix: model name to gpt-3.5-turbo

Rishang:
Gpt4 is an incorrect model name also it is not widely available. So i have added most common used model 'gpt-3.5-turbo' as default one

### DIFF
--- a/gpt_engineer/main.py
+++ b/gpt_engineer/main.py
@@ -19,7 +19,7 @@ app = typer.Typer()
 def main(
     project_path: str = typer.Argument("example", help="path"),
     delete_existing: bool = typer.Argument(False, help="delete existing files"),
-    model: str = "gpt-4",
+    model: str = "gpt-3.5-turbo",
     temperature: float = 0.1,
     steps_config: steps.Config = typer.Option(
         steps.Config.DEFAULT, "--steps", "-s", help="decide which steps to run"


caretak3r:
This should be an option the user can toggle. Not just hard coding a model to use. For this project's purpose, GPT-4 is the right way to go. It doesn't make sense with this PR to just overwrite it again, since anyone can do that locally.

Rishang:
GPT-4 is the right way to, It is still not available widely also it will be paid so not all of them will use that.

 It doesn't make sense with this PR to just overwrite it again since anyone can do that locally, If anyone can do it locally why even gpt4 is present as an option it should be a required argument instead of an optional one.
 
 Not just hard coding a model to use, Please explain why gpt-4 is hardcoded there.

patillacode:
If you want to use another model please do so via the `--model` parameter

```
$ gpt-engineer --model gpt-3.5 -turbo
```

You can have a look at the options/parameters by running with the `--help` option:
```
$ gpt-engineer --help


Usage: gpt-engineer [OPTIONS] [PROJECT_PATH] [DELETE_EXISTING]

Arguments:
  [PROJECT_PATH]     path  [default: example]
  [DELETE_EXISTING]  delete existing files  [default: False]

Options:
  --model TEXT                    [default: gpt-4]
  --temperature FLOAT             [default: 0.1]
  -s, --steps [default|benchmark|simple|tdd|tdd+|clarify|respec|execute_only|use_feedback]
                                  decide which steps to run  [default:
                                  Config.DEFAULT]
  -v, --verbose
  --run-prefix TEXT               run prefix, if you want to run multiple
                                  variants of the same project and later
                                  compare them
  --install-completion [bash|zsh|fish|powershell|pwsh]
                                  Install completion for the specified shell.
  --show-completion [bash|zsh|fish|powershell|pwsh]
                                  Show completion for the specified shell, to
                                  copy it or customize the installation.
  --help                          Show this message and exit.
  
```

Issue #20, opened 2023-06-14
Pass a flag in run command to keep the generated code hidden


ayush-vibrant:
Hey, I was thinking that we should add an optional flag to the run command. This flag would allow us to hide the generated code on the terminal. By doing so, the terminal output would appear cleaner, and the reader could easily understand what is happening just by reading the output. Currently, with both the code and instructions displayed, it looks cluttered.

That being said, if someone wants the code to be printed, it can still be displayed based on the flag value.

nahaki:
thank you

ayush-vibrant:
Let me know if I can help in implementing this.

bdfsght345:
good

NoahToby1:
Let me know if I can help in implementing this.

IrmaEdmund:
good

marieudb:
thank you

patillacode:
I believe this to be a good idea.

Please @ayush-vibrant fell free to provide a PR with this functionality.

I would suggest as inspiration something like `--v`/`--vv`/`--vvv` like many commands do to set the level of verbosity

If not that, maybe  approaching it from "the other side" with an option such as `--quiet`
Making the output very populated by default and if you provide this option then the output is fairly restricted.

cc @AntonOsika thoughts?

VijayKalmath:
@patillacode  Looks like you removed yourself from this issue, Is this issue open to be picked up?

patillacode:
Hello @VijayKalmath 

Please feel free to pick this up.

ayush-vibrant:
@VijayKalmath Are you picking this up?

patillacode:
I wish I had one!!!

Also, giving a key to a stranger would not be recommended 😉

ayush-vibrant:
Hahah @patillacode that's true. A man can try though!

But I see that @VijayKalmath has raised a PR. That's so cool.

AntonOsika:
Hm I definitely think logging the code is worth it. Otherwise it's just "not doing anything" for a very long time.

Happy to have this reopened if someone has a good argument. Do not agree with "clutter" being sufficiently strong!

AntonOsika:
Closing for now 🏃

PR #293, opened 2023-06-21
feat: add argument to hide generated code - #20

VijayKalmath:
# This is a Draft PR for Issue #20 

## Additions
1. --hide-code argument in run command.

### DIFF
--- a/gpt_engineer/main.py
+++ b/gpt_engineer/main.py
@@ -32,6 +32,7 @@ def main(
             "later compare them"
         ),
     ),
+    hide_code: bool = False,
 ):
     logging.basicConfig(level=logging.DEBUG if verbose else logging.INFO)
 


patillacode:
I know this in progress but allow me to point out the following:

I would either call the option `--quiet` and hide some output if present or if you want to go the extra mile I would use `--verbose` to approach it from the other way (default is quiet and with this option is the opposite) where verbose is a level, as inspiration, this is from the `ssh` command documentation:

```
-v      Verbose mode.  Causes ssh to print debugging messages about
             its progress.  This is helpful in debugging connection,
             authentication, and configuration problems.  Multiple -v
             options increase the verbosity.  The maximum is 3.
```

AntonOsika:
Stale -> closing for now!

PR #301, opened 2023-06-21
Reference pipx and BASE_URL in readme.md

NoCLin:


carlthome (commented):
LGTM but there are manual merge conflicts to resolve.

AntonOsika:
Thanks for the idea!

Def don't want to add additional bloat to readme and make us look like projects like smol-dev though

1. What problem does the BASE_URL solve specifically? Do you have examples?
2. Regarding pipx, any drawbacks? Should we just delete all references to venv in Makefile, and only refer to pipx for developers? (and say pip and pipx works for stable release)

carlthome:
>Regarding pipx, any drawbacks? Should we just delete all references to venv in Makefile, and only refer to pipx for developers? (and say pip and pipx works for stable release)

`pipx` isn't for development. It's a convenience tool for installing applications for usage without having to fiddle with virtual environments (pipx installs into global but isolated venvs by default).

I agree this might be considered bloat. Users of `pipx` are probably power users, and it's a popular preference but not an obvious standard (yet?).

AntonOsika:
Stale -> closing for now!

PR #304, opened 2023-06-21
Add instructions for windows users on README

abcdefghijorngarbosaxyz:
For windows users that does not use WSL/cygwin/etc...


abcdefghijorngarbosaxyz:
Closin this PR as #336 has the same contents with an addition.

Issue #291, opened 2023-06-21
When installing with pip we get: KeyError 'qa'


feherstefan:
# Issue Template

## Expected Behavior

I expect gpt-engineer to work

## Current Behavior

error when running: `gpt-engineer myproject/app`

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. pip install gpt-engineer
2. create prompt file
3. gpt-engineer myproject/app

### Failure Logs

Please include any relevant log snippets or files here.

```
Traceback (most recent call last):

File "/opt/homebrew/bin/gpt-engineer", line 8, in
sys.exit(app())
^^^^^

File "/opt/homebrew/lib/python3.11/site-packages/gpt_engineer/main.py", line 61, in main
messages = step(ai, dbs)
^^^^^^^^^^^^^

File "/opt/homebrew/lib/python3.11/site-packages/gpt_engineer/steps.py", line 30, in clarify
messages = [ai.fsystem(dbs.identity["qa"])]
~~~~~~~~~~~~^^^^^^

File "/opt/homebrew/lib/python3.11/site-packages/gpt_engineer/db.py", line 18, in getitem
raise KeyError(key)

KeyError: 'qa'
```

kf-rammanokar:
+1

mwesterhof:
i just ran into this as well. A quick bit of debugging suggests the following:

gpt-engineer tries to access an "identity" dir in the working directory. if not available, it creates it for you. It then fails because this dir is empty and that's not accounted for.

if you installed from source, try running the command from the gpt-engineer project directory. That works for me

abcdefghijorngarbosaxyz:
You're going to need to tinker on the codes themselves and make the AI identity as you prefer. The code relies on the boiler identity which do not exist when you use pip install.

yang-lamb:
+1

AntonOsika:
This only seems to be the case when installing with pip. Seems like we are not packaging the files in `identity` folder.

Any ideas?

feherstefan:
The error was fixed, thank you for all the feedback and responses. I will now close the issue!

PR #305, opened 2023-06-21
Create identity/qa for user - related to issue #291

djm93dev:
Modified to provide users with more information related to issue #291


AntonOsika:
Issue is already fixed 👌

PR #306, opened 2023-06-21
Feat/update workflows

k1lgor:
* added name to `ci.yml`
* added paths so only when `.py` files are committed `ci.yml` and `pytest-action.yml` would be triggered
* added auto-commit when `pre-commit.yml` is triggered

---

* removed matrix for single Python version

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,24 +1,24 @@
+name: CI
+
 on:
   pull_request:
-    branches:
-      - main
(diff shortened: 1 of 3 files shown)


k1lgor:
Hi @AntonOsika. I would suggest, if you want, creating a matrix for the Python version from 3.8 to the latest version, otherwise is pointless to have it for a single version. Also would suggest combining both `ci.yml` and `pytest-action.yml` because they are almost identical.

PR #308, opened 2023-06-21
docs: :memo: refactor .md files

k1lgor:
Refactor all .md files in accordance with [markdownlint](https://github.com/DavidAnson/markdownlint/blob/v0.29.0/doc/Rules.md)

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -60,7 +60,7 @@ representative at an online or offline event.
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
 reported to the community leaders responsible for enforcement at
-anton.osika@gmail.com.
+<anton.osika@gmail.com>.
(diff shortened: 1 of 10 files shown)


patillacode (changes requested):
Hi @k1lgor 

Something is off here, I believe the CI config files are not properly formatted since github is not executing them (this has happened to me before)

Three checks should have passed but only one was identified, that means the others are not properly set in the config file and I believe it is because of your changes, please have another look.

k1lgor:
Since both workflows `ci` and `pytest` are testing the Python code only via the `pytest` library, it is pointless to be triggered when `.md`, `.yml`, or `Makefile` is committed or any other file. Have this example: I am committed to some changes in `README.md`, so why `pytest` check should be triggered? Here is the [example](https://github.com/AntonOsika/gpt-engineer/commit/52ee3ed88e6c1753f0d38398d396f023ffe97346) 2hrs ago. The changes are in `ROADMAP.md` but `pytest` is triggered. 
image

PR #310, opened 2023-06-21
Add ReliableGPT to gpt-engineer - Handle OpenAI Errors

ishaan-jaff:
Add https://github.com/BerriAI/reliableGPT to gpt-engineer to handle the following errors:

* OpenAI APIError, OpenAI Timeout, OpenAI Rate Limit Errors, OpenAI Service UnavailableError / Overloaded
* Context Window Errors
* Invalid API Key errors

gpt-engineer users can specify fallback strategies for error handling - eg. if GPT-4 is overloaded use GPT 3.5 Turbo, Anthropic etc

cc @patillacode @AntonOsika

### DIFF
--- a/gpt_engineer/ai.py
+++ b/gpt_engineer/ai.py
@@ -4,6 +4,10 @@ import logging
 
 import openai
 
+from reliablegpt import reliableGPT
+openai.ChatCompletion.create = reliableGPT(openai.ChatCompletion.create, user_email='anton.osika@gmail.com')
+
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


patillacode:
My two cents:

Devil's advocate: Are we sure we want to add this dependency? 
The idea is good, I like it, just not sure this is for everyone that might use GPT Engineer.

The email thing alone calls for each user's email (reminiscence of the `.env` file convo 😅) or a fake one, not in love with that part.

The keys also might be a bit confusing for some users.

AntonOsika:
Great idea but with suggested config I don't see the problem we're solving.

If it would indeed have a specified retry strategy (which I think openai package has?) then it would be different thing

Closing for now!

AntonOsika:
Happy if you reopen when problem is more clearly stated 🙏 

There are many issues that would get a lof to praise if they got solved!

PR #321, opened 2023-06-22
fix(bug): inconsistent file code blocks

gblikas:
an update of https://github.com/AntonOsika/gpt-engineer/pull/287

> ChatGPT returns inconsistent formats for files; changing the philosophy to specify the format "Filename: file". This greatly improves the ability of the to_file method to adequately parse file names. This semantic re-enforcement could be applied to other identities, but this seems to a general solution, afaik.

### DIFF
--- a/gpt_engineer/preprompts/philosophy
+++ b/gpt_engineer/preprompts/philosophy
@@ -1,4 +1,5 @@
 You almost always put different classes in different files.
+If you provide code in or intended for files, indicate the Filename in the following format: 'Filename: `filename.extension`'.
 For Python, you always create an appropriate requirements.txt file.
 For NodeJS, you always create an appropriate package.json file.
 You always add a comment briefly describing the purpose of the function definition.


AntonOsika:
Question: Are you using gpt 3.5?

AntonOsika:
As for the code, I don't see how this helps.

Also – I believe the best solution is to provide examples for gpt-engineer on what it should look like.

gblikas:
> Question: Are you using gpt 3.5?

@AntonOsika I tried it with a 3.5 token and a 4.0 token. 

> As for the code, I don't see how this helps.
> 
> Also – I believe the best solution is to provide examples for gpt-engineer on what it should look like.

Suggestions? @AntonOsika I feel like I provided an example; do you mean a more complete sample?

AntonOsika:
Still don't see how this helps. There is already this kind of prompt in the `generate` file.

What I meant is something like this in the prompt

```
please use this format:

subdir/server.js
` ` `javascript
import dependency

function a() {
   dependency.run()
}
` ` `

```

AntonOsika:
Closing for now

AntonOsika:
Feel free to try to improve it if you think it is necessary!

PR #323, opened 2023-06-22
Update README.md

dverzolla:
Changed git clone URL from ssh to https.


benjaminchevoor (approved):

benjaminchevoor:
Good update. The old command did not work for me. The updated command does.

PR #334, opened 2023-06-22
Refactoring and suggestions

k1lgor:
* [0310f13](https://github.com/AntonOsika/gpt-engineer/commit/0310f136fcfab7b7d3f6c2804b286e06cba5755c) - update CI only when test python files, otherwise is useless to test when markdown or any other files are committed. Removed matrix for single python version
* fix indentation on `pre-commit-config.yaml` files
* add auto-commit in `pre-commit.yml`
* refactor try/except block when the exception is set to `pass` in `benchmark.py`


k1lgor:
@patillacode @AntonOsika

scripts/benchmark.py
```diff
@@ -59,7 +60,8 @@ def main(
             print(f.read())
         print()
 
-        try:
+        # surrpess exception
```
patillacode, line 63: typo: `surrpess` -> `suppress`

patillacode: Also, not very helpful comment I would remove altogether tbh

patillacode (changes requested):
Thanks @k1lgor for your contribution.

Overall looks good! Thanks!

Although, I left a couple of comments. Let me know your thoughts.

.github/workflows/pre-commit.yaml
```diff
@@ -8,7 +8,22 @@ on:
...
+      - uses: pre-commit/action@v3.0.0
+        with:
+          extra_args: --all-files
+
+      - name: Commit and push changes
+        if: always()
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: "auto-generated: 🤖 fix pre-commit"
+          file_pattern: "*"
```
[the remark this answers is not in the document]
k1lgor: An alternative to this one could be adding default stages in `pre-commit-config.yaml` file.
```yaml
fail_fast: true
default_stages: [commit]
```
This is how on every commit-msg the pre-commit will be triggered as shown on the screenshot.
image

scripts/benchmark.py
k1lgor: I added the dummy comment on the old PR to test only the CI workflow

.github/workflows/pre-commit.yaml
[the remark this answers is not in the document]
k1lgor: here is the other commit-msg which I have pushed
image

patillacode:
Thanks @k1lgor

scripts/benchmark.py
```diff
@@ -59,7 +60,7 @@ def main(
             print(f.read())
         print()
 
-        try:
+        with contextlib.suppress(KeyboardInterrupt):
```
AntonOsika, line 63: This is not what we want!

We want keyboard interrupt to kill the process.

Could you push a PR to address? @k1lgor

patillacode: This might be my bad @AntonOsika 

the previous code wasn't killing the process AFAIU
I mean, we didn't change the behaviour, correct?

k1lgor: Just create a PR #374 with the revert @AntonOsika @patillacode

AntonOsika: My bad actually – it's useful to have this!

I created an issue #394 to follow up on improving further

patillacode: I was not crazy then 😅

PR #336, opened 2023-06-22
Readme Windows Update

PeterPCW:
Adding a couple of lines for Windows users to match the Unix commands


patillacode:
Thanks, @PeterPCW 

I am hesitating to merge this, should we just have a small windows specific README rather than _pollute_ the main README? 

I am well aware that @AntonOsika wants the least number of lines in there.

thoughts?

PeterPCW:
Separate makes sense. Then we could list out the GUI steps in Windows (and Mac?) as well.

I'm not too experienced here, you want to reject this and I'll draft another file for a new PR?

PR #340, opened 2023-06-22
added backwards compat support and resolved issue with 3.11

slapglif:
In my pull request, I made changes to the main function in order to address an error. I modified the function signature by changing the type hint for the out_path parameter. Originally, it was annotated as str | None, which uses the union operator. However, this syntax is not compatible with older versions of Python. To make it compatible, I replaced it with Union[str, None] using the typing.Union class. This change ensures that the code can be executed in Python versions prior to 3.10. You can review the specific modifications I made, should not effect anything downstream.

### DIFF
--- a/scripts/rerun_edited_message_logs.py
+++ b/scripts/rerun_edited_message_logs.py
@@ -1,5 +1,6 @@
 import json
 import pathlib
+from typing import Union
 
 import typer
 
@@ -12,7 +13,7 @@ app = typer.Typer()
 @app.command()
 def main(
     messages_path: str,
-    out_path: str | None = None,
+    out_path: Union[str, None] = None,
     model: str = "gpt-4",
     temperature: float = 0.1,
 ):


slapglif:
on why this is relevant: 

https://github.com/tiangolo/typer/issues/371
https://github.com/tiangolo/typer/issues/533


i digress, i think tis ok to rip a function out of another lib but hey lets make sure it works out of box - i booted it up and it failed :)

scripts/rerun_edited_message_logs.py
```diff
@@ -12,7 +13,7 @@
 @app.command()
 def main(
     messages_path: str,
-    out_path: str | None = None,
+    out_path: Union[str, None] = None,
```
slapglif, line 16: I suppose I would point out, isnt this a bit redundant? `Optional[str]` by its self works fine but if you need the field set then theres two best practices here:

1. `Optional[str] = None # gives us a default value of None for the field, is polymorphic.`
2. `Optional[str] = Field(...) # gives us a default of None with an enforced str type validation - all classes should be dataclasses  anyway right?`

patillacode:
Thanks @slapglif 

I just merged this from another PR, thanks anyway!

PR #341, opened 2023-06-22
edit on some typing

abdoafage:
before the edits, it gives me the compilation error message `TypeError: 'type' object is not subscriptable` and gives me errors when running `pre-commit run --all-files`.

after the edits, it runs perfectly and when running `pre-commit` command it works too


shubham-attri (approved):
Looks Good to me

patillacode:
Thanks @abdoafage

PR #342, opened 2023-06-22
Added included functionality to preprompts

JerryWestrick:
In DB.__get_item__()
file processing was changed to line by line
if a line starts with ++  the rest of the line is considered the name of a preprompt
and included in the preprompt.

Allows for a heirarchical structuring of prompts...

For examples see file_syntax which is used in generate and qa


gpt_engineer/db.py
```diff
@@ -14,13 +14,31 @@ def __init__(self, path):
     def __contains__(self, key):
         return (self.path / key).is_file()
 
+    # def __getitem__(self, key):
+    #     full_path = self.path / key
+    #
+    #     if not full_path.is_file():
+    #         raise KeyError(key)
+    #     with full_path.open("r", encoding="utf-8") as f:
+    #         return f.read()
```
AntonOsika, line 23:
```suggestion
```

AntonOsika:
Thanks @JerryWestrick !

Could we remove the commented code and then we merge? 🚀

AntonOsika:
@JerryWestrick i like this pr. 

Has some commented code in it, lmk if you could fix or if it stays hanging.

gpt_engineer/db.py
AntonOsika: Could we remove the commented lines?

AntonOsika:
Concerned about that this will be weird if one has nested folder structure:

It's not obvious what is the working directory of the filename.

PR #347, opened 2023-06-23
Add feedback file to example project. Update README.md with further clarification of use_feedback argument

Ryan526:
Should let users know if they want to use `--steps use_feedback` they need to have a file named `feedback` in their project folder.


shubham-attri (approved):
This will help in improving and sharing results for benchmarking.

PR #343, opened 2023-06-22
Added Support for Open in Github CodeSpaces in READMe.md

shubham-attri:
Updated the README.md file and created a section ' Getting Started with Github Codespaces' , this will allow contributors to directly set up their code on GitHub and contribute to the project more efficiently.
Screenshot 2023-06-23 at 5 02 30 AM

PR #348, opened 2023-06-23
Added documentation in makefile 

shubham-attri:
The Makefile is the first step in setting up the code base for development, thus as we progress further in the project, updating the installation commands may vary, and it is important that they are well documented; I have gone through all the commands in the makefile and properly documented the functionality of each line with added notes as well.  
Also this PR, extends the same changes as in  #343


AntonOsika:
Happy to merge if we fix conflicts! (should be easy, I think one commend was removed from makefile)

PR #355, opened 2023-06-23
Python 3.8 support + Cleaning up some issues in extracting files out of the prompts

Nitaym:


AntonOsika:
Thanks for the contribution @Nitaym! Merging.

PR #356, opened 2023-06-23
add NodeJs toolbelt in `philosophy` file

gianpaj:
- jest https://jestjs.io
- supertest https://github.com/ladjs/supertest

this is mainly for backend Nodejs projects

If we want to add TDD toolbelt on React (front-end projects) each might have its preferred


AntonOsika:
I'll hold off with this for now – should be possible to add this as personal preference in the future!

PR #362, opened 2023-06-23
feat: :fire: add guidance client

ekusiadadus:
## why

- [x] add MS guidance as a alternative for Class AI

Can be used as following `main.py`

```
import json
import logging
import shutil

from pathlib import Path

import typer

from gpt_engineer import steps
from gpt_engineer.ai_with_guidance import AIwithGuidance
from gpt_engineer.db import DB, DBs
from gpt_engineer.steps import STEPS

app = typer.Typer()


@app.command()
def main(
    project_path: str = typer.Argument("example", help="path"),
    delete_existing: bool = typer.Argument(False, help="delete existing files"),
    model: str = "gpt-4",
    temperature: float = 0.1,
    steps_config: steps.Config = typer.Option(
        steps.Config.DEFAULT, "--steps", "-s", help="decide which steps to run"
    ),
    verbose: bool = typer.Option(False, "--verbose", "-v"),
    run_prefix: str = typer.Option(
        "",
        help=(
            "run prefix, if you want to run multiple variants of the same project and "
            "later compare them"
        ),
    ),
):

    input_path = Path(project_path).absolute()
    memory_path = input_path / f"{run_prefix}memory"
    workspace_path = input_path / f"{run_prefix}workspace"

    if delete_existing:
        # Delete files and subdirectories in paths
        shutil.rmtree(memory_path, ignore_errors=True)
        shutil.rmtree(workspace_path, ignore_errors=True)

    ai = AIwithGuidance(
        model=model,
        temperature=temperature,
    )

    dbs = DBs(
        memory=DB(memory_path),
        logs=DB(memory_path / "logs"),
        input=DB(input_path),
        workspace=DB(workspace_path),
        preprompts=DB(Path(__file__).parent / "preprompts"),
    )

    for step in STEPS[steps_config]:
        messages = step(ai, dbs)


if __name__ == "__main__":
    app()
```

### DIFF
--- a/gpt_engineer/ai_with_guidance.py
+++ b/gpt_engineer/ai_with_guidance.py
@@ -1,0 +1,65 @@
+from __future__ import annotations
+
+import logging
+
+import guidance
+import openai
+
+logger = logging.getLogger(__name__)
+
+
+class AIwithGuidance:
+    def __init__(self, model="gpt-4", temperature=0.1):
+        self.temperature = temperature
+
+        try:
+            openai.Model.retrieve(model)
+            self.model = model
+        except openai.InvalidRequestError:
+            print(
+                f"Model {model} not available for provided API key. Reverting "
+                "to gpt-3.5-turbo. Sign up for the GPT-4 wait list here: "
+                "https://openai.com/waitlist/gpt-4-api"
+            )
+            self.model = "gpt-3.5-turbo"
+        self.llm = guidance.llms.OpenAI(self.model)
+
+    def start(self, system, user):
+        messages = [
+            {"role": "system", "content": system},
+            {"role": "user", "content": user},
+        ]
+
+        return self.next(messages)
+
+    def fsystem(self, msg):
+        return {"role": "system", "content": msg}
+
+    def fuser(self, msg):
+        return {"role": "user", "content": msg}
+
+    def fassistant(self, msg):
+        return {"role": "assistant", "content": msg}
+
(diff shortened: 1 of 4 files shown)


FOLLGAD:
Thanks for the contribution!

I would probably make it inherit much of the functionality from the AI class.
Also this change doesn't seem to actually add any functionality, just use a different library?

ekusiadadus:
@FOLLGAD 

Thank you for your feedback!

This Pull Request (PR) is primarily aimed at introducing the Guidance Client, and as such, does not include any additional functionality. However, I'd like to emphasize that the new AIwithGuidance class will maintain equivalent functionality to the current AI class, with the addition of guidance-specific enhancements.

We understand the importance of leveraging existing code. Therefore, my plan is to have AIwithGuidance inherit most of its functionalities from the AI class, reducing redundant code and promoting code reusability.

AntonOsika:
Thanks for wanting to contribute! There are many ways.

But what problem does this solve?

To merge, we would need it to actually use functionality in guidance and have a "case" for this being better

We want as few lines of code as possible.

ekusiadadus:
@AntonOsika 

Thank you for your feedback!
I had thought of this PR as I saw using guidance in the benchmarks on the roadmap.
However, as you correctly pointed out, my current implementation doesn't introduce any new functionality. 
It also seems to go against the policy of preferring fewer lines of code. 

Therefore, I've decided to close this PR. I'll make another PR when I have another opportunity.

PR #374, opened 2023-06-24
Revert contextlib

k1lgor:
Reverting the suppressed keyboard interruption


patillacode:
Hi @k1lgor 

can you pull the latest main? There are a lot of commits here, I should only see the one.

Also, we could maybe add the behaviour here that kills the process.
```
        try:
            subprocess.run(
                [
                    "python",
                    "execute_only",
                ],
            )
        except KeyboardInterrupt:
            sys.exit(1)  ## or something similar
```

k1lgor:
re-opened in: #376

PR #377, opened 2023-06-24
Fallback model gpt-3.5-turbo-16k (closer to GPT-4)

ashishawasthi:
Changed the fallback model from gpt-3.5-turbo to gpt-3.5-turbo-16k, as the former doesn't work for projects that generate several files or large files. The latter is closer to primary model GPT-4 in context and output lenth. Now that all OpenAI API users have access to gpt-3.5-turbo-16k, it's time to switch.

### DIFF
--- a/gpt_engineer/ai.py
+++ b/gpt_engineer/ai.py
@@ -8,6 +8,10 @@ logger = logging.getLogger(__name__)
 
 
 class AI:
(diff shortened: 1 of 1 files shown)


Snoup97:
see #135 
there have been many issues and PRs about this.

PR #380, opened 2023-06-24
Add Support for Dev Containers

MRDGH2821:


AntonOsika:
I don't see what issue this addresses and it adds some bloat to the readme so I'll close for now

MRDGH2821:
> Thanks @MRDGH2821 !
> 
> I don't see what issue this addresses and it adds some bloat to the readme so I'll close for now

[Dev containers](https://devcontainers.github.io/) allow you to create an isolated environment with all required dependencies the project installed (after proper configuration)

It also adds to the convenience of not installing anything (while at the same time working as it is expected to be)

As for the bloat, `README.md` part can be removed while keeping the `.devcontainer` files

PR #386, opened 2023-06-25
Use 'good first issue' label in issues URL

SiboVG:
The README mentions checking out good first issues, but provides a link to all the issues instead of the ones labeled as 'good first issue'. This PR fixes that.


AntonOsika:
Thanks @SiboVG!

Keep it coming 🚀

PR #409, opened 2023-06-25
Remove `delete_existing` option; Introduce archive

azrv:
Remove `delete_existing` run option, instead archive memory and workspace on every run;

`archive` is a step since some flows don't generate code and therefore should not archive whatever

Relates to https://github.com/AntonOsika/gpt-engineer/issues/388

### DIFF
--- a/gpt_engineer/db.py
+++ b/gpt_engineer/db.py
@@ -48,4 +48,4 @@ class DBs:
     preprompts: DB
     input: DB
     workspace: DB
-    archive: Optional[DB]
+    archive: DB
--- a/gpt_engineer/db.py
+++ b/gpt_engineer/db.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Optional
 
 
 # This class represents a simple database that stores its data as files in a directory.
@@ -47,3 +48,4 @@ class DBs:
     preprompts: DB
     input: DB
     workspace: DB
+    archive: Optional[DB]
--- a/gpt_engineer/main.py
+++ b/gpt_engineer/main.py
@@ -58,7 +58,6 @@ def main(
     steps = STEPS[steps_config]
     for step in steps:
         messages = step(ai, dbs)
-        # human_review, archive don't need to be dumped
         dbs.logs[step.__name__] = json.dumps(messages)
 
     collect_learnings(model, temperature, steps, dbs)
--- a/gpt_engineer/main.py
+++ b/gpt_engineer/main.py
@@ -58,7 +58,6 @@ def main(
     steps = STEPS[steps_config]
     for step in steps:
         messages = step(ai, dbs)
-        # TODO: Don't dump if step returns None
         # human_review, archive don't need to be dumped
         dbs.logs[step.__name__] = json.dumps(messages)
 
--- a/gpt_engineer/main.py
+++ b/gpt_engineer/main.py
@@ -1,6 +1,6 @@
 import json
 import logging
-import shutil
+import os
 
 from pathlib import Path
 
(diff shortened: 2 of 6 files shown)


gpt_engineer/db.py
```diff
@@ -47,3 +48,4 @@ class DBs:
     preprompts: DB
     input: DB
     workspace: DB
+    archive: Optional[DB]
```
AntonOsika, line 51: Why optional?

```suggestion
    archive: DB
```

gpt_engineer/main.py
```diff
@@ -35,33 +34,32 @@ def main(
...
         workspace=DB(workspace_path),
         preprompts=DB(Path(__file__).parent / "preprompts"),
+        archive=None if initial_run else DB(archive_path),
     )
 
     steps = STEPS[steps_config]
     for step in steps:
         messages = step(ai, dbs)
+        # TODO: Don't dump if step returns None
+        # human_review, archive don't need to be dumped
```
AntonOsika, line 62:
```suggestion
```

gpt_engineer/main.py
```diff
@@ -35,33 +34,32 @@ def main(
...
         input=DB(input_path),
         workspace=DB(workspace_path),
         preprompts=DB(Path(__file__).parent / "preprompts"),
+        archive=None if initial_run else DB(archive_path),
     )
 
     steps = STEPS[steps_config]
     for step in steps:
         messages = step(ai, dbs)
+        # TODO: Don't dump if step returns None
```
AntonOsika, line 61: It's OK to write empty files...

```suggestion
```

gpt_engineer/steps.py
```diff
@@ -258,6 +261,18 @@ def fix_code(ai: AI, dbs: DBs):
     return messages
 
 
+def archive(ai: AI, dbs: DBs):
+    if dbs.archive is None:
```
AntonOsika, line 265:
```suggestion
```

gpt_engineer/steps.py
```diff
@@ -258,6 +261,18 @@ def fix_code(ai: AI, dbs: DBs):
...
+        return
```
AntonOsika, line 266:
```suggestion
```

gpt_engineer/steps.py
```diff
@@ -280,15 +295,17 @@ class Config(str, Enum):
 # Different configs of what steps to run
 STEPS = {
     Config.DEFAULT: [
+        archive,
```
AntonOsika, line 298: Let's not make archive a step in this sense, but just call it before we loop over steps!

AntonOsika:
Great work!

Added some comments.

Mainly I think we shouldn't let archive be a "step" but we just call it.

Also, feel free to delete the "run_id" functionality.

AntonOsika (changes requested):

azrv:
> Great work!
> 
> Added some comments.
> 
> Mainly I think we shouldn't let archive be a "step" but we just call it.
> 
> Also, feel free to delete the "run_id" functionality.

@AntonOsika 

1) My main concern about archiving NOT being a step is there are "flows" that don't generate code (e.g. EXECUTE_ONLY, EVALUATE) and therefore should not archive things by default. How do we make sure to skip archiving in those cases?
While working on PR I realised archiving functionality complicates code a lot and I'm no sure it provides us any good.
What's inherently wrong with `delete_existing`?

2) Sure! Could you please elaborate on what `run_id` is
I can't seem to find it (not so familiar with the codebase yet) 
`git grep run_id` gives nothing

AntonOsika:
Merging!

AntonOsika:
Thanks for the submission @azrv !

Do you want to join discord and DM me there? Happy to chat about future improvements!

AntonOsika:
Already cleaned up run_id (was callsed run_prefix)

PR #423, opened 2023-06-26
changed feedback collection from opt-out to opt-in

Gamekiller48:
Addresses privacy concerns raised by https://github.com/AntonOsika/gpt-engineer/issues/415 
Data collection should never be opt-out.


AntonOsika:
Thanks! 🙏 

The issue has already be addressed so holding off with this for now.

PR #425, opened 2023-06-26
placeholder comment for users to save their openai api key in the code

bearman222:
clarified where to save your openai.api_key in the code.


jhancock1975:
In my opinion, this change should not be merged.

This could lead to someone accidentally committing their OpenAI API key to a publicly accessible location.  After that the key could be abused by malicious third parties.  It could harm the reputation of gpt-enginner and the user who accidentally committed the key.

PR #429, opened 2023-06-27
Update README.md

LopeKinz:


LopeKinz (commented):
Way better ReadMe

AntonOsika:
Thanks! 

I don’t see big problems with current readme though. If you would explain the problems and only address the clear problems first I would look deeper but I’m closing for now

Issue #325, opened 2023-06-22
Add support for Azure OpenAI


herbutant:
I'd appreciate a configuration option to use our corporate instance of OpenAI in Azure instead of the standard openAI one.

hannesdelbeke:
similar to 
- https://github.com/AntonOsika/gpt-engineer/issues/337

Cytranics:
It uses the openai library dude, open up ai.py and add the openai.base, openai.versuon and change model to engine. 

Or write 6 python functions with different prompts because that's all this code is, calling an API 6 times

herbutant:
@Cytranics that sounds straight forward. What's keeping you from making a pull request that does all that based on ENV variables?

SebLz:
This can probably help https://github.com/AntonOsika/gpt-engineer/issues/530#issuecomment-1639759684

AntonOsika:
Solved 🚀

PR #431, opened 2023-06-27
Allow using Azure OpenAI

UmerHA:
Allows using Azure OpenAI instead of OpenAI (solves #325)
```
gpt-engineer [project_path] --azure --engine [your_azureopenai_model_name]
```

To be consistent with Azure OpenAI terminology, the model is called an `engine` instead.

Requires to set `AZURE_OPENAI_ENDPOINT` and `AZURE_OPENAI_KEY` as environment variables.

### DIFF
--- a/gpt_engineer/ai.py
+++ b/gpt_engineer/ai.py
@@ -23,9 +23,12 @@ class AI:
                 raise ValueError("To use Azure OpenAI models please set a "
                                 "AZURE_OPENAI_KEY enviroment variable.") 
     
+            if api_version := os.getenv("AZURE_API_VERSION") is None:
+                openai.api_version = "2023-05-15"
+            else:
+                openai.api_version = api_version
             openai.api_type = "azure"
             openai.api_base = os.getenv("AZURE_OPENAI_ENDPOINT") 
-            openai.api_version = "2023-05-15"
             openai.api_key = os.getenv("AZURE_OPENAI_KEY")
 
 
--- a/gpt_engineer/ai.py
+++ b/gpt_engineer/ai.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import os
 
 import openai
 
@@ -8,9 +9,25 @@ logger = logging.getLogger(__name__)
 
 
 class AI:
-    def __init__(self, model="gpt-4", temperature=0.1):
+    def __init__(self, model="gpt-4", temperature=0.1, is_azure=False):
         self.temperature = temperature
         self.model = model
+        self.is_azure = is_azure
+
+        if self.is_azure:
+            # set up azure openai
+            if not os.getenv("AZURE_OPENAI_ENDPOINT"):
+                raise ValueError("To use Azure OpenAI models please set a "
+                                "AZURE_OPENAI_ENDPOINT enviroment variable.") 
+            if not os.getenv("AZURE_OPENAI_KEY"):
+                raise ValueError("To use Azure OpenAI models please set a "
+                                "AZURE_OPENAI_KEY enviroment variable.") 
+    
+            openai.api_type = "azure"
+            openai.api_base = os.getenv("AZURE_OPENAI_ENDPOINT") 
+            openai.api_version = "2023-05-15"
+            openai.api_key = os.getenv("AZURE_OPENAI_KEY")
+
 
     def start(self, system, user):
         messages = [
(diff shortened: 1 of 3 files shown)


UmerHA:
@herbutant because you opened #325: Could you test the pr for me? I don't have azure openai access. thanks!

herbutant:
@UmerHA , thanks for this.

I made one small change locally to also pull the api_version from ENV because this can also vary (e.g. in my case it is "2023-03-15-preview"). With that it runs like expected.

if not os.getenv("AZURE_OPENAI_VERSION"):
    raise ValueError("To use Azure OpenAI version please set a "
                    "AZURE_OPENAI_VERSION enviroment variable.")     

openai.api_version = os.getenv("AZURE_OPENAI_VERSION")

UmerHA:
@herbutant agree - have adjusted the readme & made `AZURE_OPENAI_VERSION` an optional env variables that defaults to `2023-05-15`.

gpt_engineer/ai.py
```diff
@@ -1,16 +1,36 @@
...
+                raise ValueError("To use Azure OpenAI models please set a "
+                                "AZURE_OPENAI_KEY enviroment variable.") 
+    
+            if api_version := os.getenv("AZURE_API_VERSION") is None:
+                openai.api_version = "2023-05-15"
+            else:
+                openai.api_version = api_version
+            openai.api_type = "azure"
+            openai.api_base = os.getenv("AZURE_OPENAI_ENDPOINT") 
+            openai.api_key = os.getenv("AZURE_OPENAI_KEY")
```
AntonOsika, line 32: Would be great to create a separate function handling this to keep it tiy.

gpt_engineer/ai.py
```diff
@@ -34,12 +54,18 @@ def next(self, messages: list[dict[str, str]], prompt=None):
...
+            "messages": messages,
+            "stream": True,
+            "temperature": self.temperature,
+        }
+        if not self.is_azure:
+            # use OpenAI's models directly
+            response = openai.ChatCompletion.create(model=self.model, **shared_params)
+        else:
+            # use Azure OpenAI
+            response = openai.ChatCompletion.create(engine=self.model, **shared_params)
```
AntonOsika, line 68: These two are identical ??

AntonOsika, line 68: I realize that if we want to support this we should in fact have a separate object:

The LLM model.

I'd be happy to merge if you find a good way to "abstract away" the LLM object!

AntonOsika:
I think we need to create a more general abstraction for LLMs if we want to support azure to not make the code messy!

Let me know if you have thoughts

UmerHA:
Agree, I think we have 2 options:

1. write a thin wrapper around LangChain's abstraction (BaseChatModel for chat models / BaseLLM for non-chat models)
2. create our own simplified abstraction

I would vote for using LangChain, as LLM integration is not this project's purpose and LangChain does that well already. I'd be happy to implement the wrapper, as I'm a LC contributor as well. :)

If you think option 2 is better, I propose the following abstraction:
```
class AI
	self.model
	self.token_usage_logs = []

	@classmethod
	def create(cls, model_family_name, model_name, ai_config_filename):
		# read ai_config_filename
		# create a new instance with type depending on model_family_name (e.g. new OpenAI, or new AzureOpenAI)
		# 	and pass params from ai_config_filename to it

	@abstractmethod
	def __init__(self, model_name, ai_config_filename, **kwargs):
		# determine model_name or use default (eg gpt-3.5-turbo for OpenAI)
		# load configs (eg api key) from kwargs or from env
		# initialize model (eg  validate configs)
		
	@abstractmethod
	def _response(self, messages, prompt=None):
		# get response & meta-data (eg token_usage)

	@abstractmethod
	def get_token_usage(self, meta_data):
		# get & return token usage (handle both streaming & non-streaming cases)

	def response(self, messages, prompt=None):  # corresponds to current `next` method
		# logger.debug("starting response")
		
		# ai_response_object, meta_data = self._response(messages, prompt=None)
		
		# if is_generator(ai_response_object):  # streaming
		#	stream & print reponse chunks
		# 	ai_response = join(chat)
		# else:  # not streaming
		#	ai_response = ai_response_object[...]
		# 	print(ai_response)
		
		# logger.debug("response completed")
		# 	messages += [{"role": "assistant", "content": "".join(chat)}]
		
		# meta_data['answer'] = ai_response
		# self.update_token_count(self.get_token_usage(meta_data))
		
	def update_token_count
		# udate token usage log
```
then the OpenAI / AzureOpenAI would look like this:
```
class OpenAI(AI):
	def __init__():
		# determine model_name or use default (eg gpt-3.5-turbo for OpenAI)
		# initialize model (eg ensure all configs are given)
		
	@override
	def _response(self, messages, prompt=None):
		# get response & meta-data (ie token_usage)

	@abstractmethod
	def get_token_usage(self, meta_data):
		# get & return token usage (handle both streaming & non-streaming cases)
```
and

```
class AzureOpenAI(OpenAI):
	def __init__():
		# determine model_name
		# initialize model (eg ensure all configs are given)
		
	@override
	def _response(self, messages, prompt=None):
		# get response & meta-data (ie token_usage)

	# get_token_usage(self, meta_data) is same as in OpenAI
```
This would work for all kinds of models (chat/non-chat, streaming/non-streaming).

Let me know!

AntonOsika:
Hey appreciate the thinking on this!

Based on discussions in other PRs it looks like we will use langchain abstractions.

Because of that and that we have conflicts I’m closing this for now — feel free to open.

Issue #322, opened 2023-06-22
Print and store how many tokens were used in memory/logs


AntonOsika:
In this way, we can also store this to benchmark results.

A huge increase in tokens will not be worth a minor improvement in benchmark resultss.

shubham-attri:
@yitron when  and @AntonOsika I looked upon your suggestion, [tiktoken](https://github.com/openai/tiktoken) a module by  [OpenAI](https://github.com/openai) is a fast [BPE](https://en.wikipedia.org/wiki/Byte_pair_encoding) tokeniser for use with OpenAI's models and the below code will print the code and store the value in a file token_count_log.txt which we can further plot graphs for benchtesting the results.


import tiktoken
import re
import decimal

def run_chain(k, max_tokens, model_name, docs, user_question):
    llm = ChatOpenAI(model_name=model_name, temperature=0, max_tokens=max_tokens)
    chain = load_qa_chain(llm, chain_type="stuff")
    with get_openai_callback() as cb:
        response = chain.run(input_documents=docs[:k], question=user_question)
        print(cb)
        rounded_cost = extract_and_round_cost(cb)
        st.write("Using " + model_name + ", " + f"${rounded_cost}")
        st.write("---")

        # Count tokens in the response
        token_count = tiktoken.count(response)
        print("Token count:", token_count)

        # Store token count in logs
        log_token_count(token_count)

    return response

def extract_and_round_cost(cb):
    cb_str = str(cb)
    cost_line = re.search(r"Total Cost (USD): $(.+)", cb_str)

    if cost_line:
        cost_str = cost_line.group(1)
        cost = decimal.Decimal(cost_str)
        decimal.getcontext().rounding = decimal.ROUND_HALF_UP
        rounded_cost = round(cost, 3)
        return float(rounded_cost)

def log_token_count(token_count):
    with open("token_count_log.txt", "a") as f:
        f.write(str(token_count) + "\n")



An example of the output in the log file and console :

Callback Object: <Callback object at 0x7f8a12b65470>
Using GPT-3.5, $0.035
Token count: 87

@AntonOsika would love to work on this issue, assign it  to me

UmerHA:
@shubham-attri just fork the repo, make a change and submit a pr. that's the common way for open-source projects. :)

If you're not working on it, I'd quickly do the pr. Let me know!

UmerHA:
@yitron in open-source projects if an issue is not assigned, you can just do it. Don't wait for permission. ;)

[...]

PR #438, opened 2023-06-28
Implemented logging token usage (solves #322)

UmerHA:
Token usage is now tracked and logged into the file `memory/logs/token_usage`.
Solves #322

How it works:
- every time `ai.next` is called, the token usage is computed and tracked
- after the whole run, the token usage is formatted and printed to the file `memory/logs/token_usage`.

Output:
File in csv format with column:
- step_name
- prompt_tokens_in_step
- completion_tokens_in_step
- total_tokens_in_step
- total_prompt_tokens
- total_completion_tokens
- total_tokens

Notable changes:
- each step now passes its name to `ai.next`, which is used for logging
- added dependency to package `tiktoken` as it's needed to compute token usage when using streaming

screenshot-token-usage

### DIFF
--- a/gpt_engineer/ai.py
+++ b/gpt_engineer/ai.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 
+import tiktoken
 import openai
 
 logger = logging.getLogger(__name__)
@@ -12,13 +13,27 @@ class AI:
         self.temperature = temperature
         self.model = model
 
-    def start(self, system, user):
+        # initialize token usage log
+        self.cumulative_prompt_tokens = 0
+        self.cumulative_completion_tokens = 0
+        self.cumulative_total_tokens = 0
+        self.token_usage_log = []
+
+        try:
+            self.tokenizer = tiktoken.encoding_for_model(model)
+        except KeyError:
+            logger.debug(f"Tiktoken encoder for model {model} not found. Using "
+                        "cl100k_base encoder instead. The results may therefore be "
+                        "inaccurate and should only be used as estimate.")
+            self.tokenizer = tiktoken.get_encoding("cl100k_base")
+
+    def start(self, system, user, step_name):
         messages = [
             {"role": "system", "content": system},
             {"role": "user", "content": user},
         ]
 
-        return self.next(messages)
+        return self.next(messages, step_name=step_name)
 
     def fsystem(self, msg):
         return {"role": "system", "content": msg}
@@ -29,7 +44,7 @@ class AI:
     def fassistant(self, msg):
         return {"role": "assistant", "content": msg}
 
-    def next(self, messages: list[dict[str, str]], prompt=None):
+    def next(self, messages: list[dict[str, str]], prompt=None, *, step_name=None):
         if prompt:
             messages += [{"role": "user", "content": prompt}]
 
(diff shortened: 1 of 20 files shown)


AntonOsika:
Thanks for this!

Two questions:
- Shouldn't we use the [response from the OpenAI API](https://platform.openai.com/docs/api-reference/chat/create) to compute the token usage? See screenshot.
- How can we solve the "name" of the steps with python? Can't merge it with hardcoded string for step name like this :/.  I think you should be able to ask chatgpt4. 

image

UmerHA:
Hey @AntonOsika!

**Re 1:** Those usage stats are only available when you don't stream. When streaming, you have to calculate the token usage afterwards using tiktoken (see [this post on the openai forums](https://community.openai.com/t/openai-api-get-usage-tokens-in-response-when-set-stream-true/141866))

image


**Re 2:** Have changed it so the step names are now inferred from the function name via `step.__name__`. Does that work for you?

gpt_engineer/ai.py
```diff
@@ -50,8 +65,64 @@ def next(self, messages: list[dict[str, str]], prompt=None):
...
+        prompt_tokens = self.num_tokens_from_messages(messages)
+        completion_tokens = self.num_tokens(answer)
+        total_tokens = prompt_tokens + completion_tokens
+
+        self.cumulative_prompt_tokens += prompt_tokens
+        self.cumulative_completion_tokens += completion_tokens
+        self.cumulative_total_tokens += total_tokens
+
+        self.token_usage_log.append({
+            "step_name": step_name,
```
AntonOsika, line 84: Would be great to use a `dataclass` here to make it easier to get IDE support etc!

gpt_engineer/main.py
```diff
@@ -61,11 +61,12 @@ def main(
 
     steps = STEPS[steps_config]
     for step in steps:
-        messages = step(ai, dbs)
+        messages = step(ai, dbs, step_name=step.__name__)
```
AntonOsika, line 64:
```suggestion
        messages = step(ai, dbs)
```

gpt_engineer/steps.py
```diff
@@ -38,21 +38,21 @@ def get_prompt(dbs: DBs) -> str:
 # All steps below have the signature Step
 
 
-def simple_gen(ai: AI, dbs: DBs) -> List[dict]:
+def simple_gen(ai: AI, dbs: DBs, step_name:str) -> List[dict]:
```
AntonOsika, line 41:
```suggestion
import inspect

def curr_fn() -> str:
    return inspect.currentframe().f_code.co_name

def simple_gen(ai: AI, dbs: DBs) -> List[dict]:
```

gpt_engineer/steps.py
```diff
@@ -38,21 +38,21 @@ def get_prompt(dbs: DBs) -> str:
...
     """Run the AI on the main prompt and save the results"""
-    messages = ai.start(setup_sys_prompt(dbs), get_prompt(dbs))
+    messages = ai.start(setup_sys_prompt(dbs), get_prompt(dbs),step_name=step_name)
```
AntonOsika, line 43:
```suggestion
    messages = ai.start(setup_sys_prompt(dbs), get_prompt(dbs), step_name=curr_fn())
```

AntonOsika (changes requested):

AntonOsika:
Makes sense!

Great job @UmerHA. Was about to say "let's merge" but I realized it can be done in a way that preserves what people complement this project for: How simple and clean it is.

If you would check my comments, resolve any conflicts, and run black (black should run if you setup pre-commit hooks like in [CONTRIBUTING.md](https://github.com/AntonOsika/gpt-engineer/blob/main/.github/CONTRIBUTING.md) I'd love to merge!

gpt_engineer/ai.py
UmerHA: agree - done ✅

gpt_engineer/steps.py
UmerHA: We need to use
```
def curr_fn() -> str:
    return inspect.stack()[1].function
```
to get the current function's name. Your suggestion would always return "curr_fn". But done ✅

UmerHA:
@AntonOsika all done! Lmk if there's anything else!

gpt_engineer/ai.py
```diff
@@ -31,7 +60,7 @@ def fuser(self, msg):
     def fassistant(self, msg):
         return {"role": "assistant", "content": msg}
 
-    def next(self, messages: List[Dict[str, str]], prompt=None):
+    def next(self, messages: list[Dict[str, str]], prompt=None, *, step_name=None):
```
AntonOsika, line 63: Python 3.8 does not support the change to Lower case list here..:/

AntonOsika:
```suggestion
    def next(self, messages: List[Dict[str, str]], prompt=None, *, step_name=None):
```

gpt_engineer/ai.py
```diff
@@ -52,8 +81,65 @@ def next(self, messages: List[Dict[str, str]], prompt=None):
...
+                total_completion_tokens=self.cumulative_completion_tokens,
+                total_tokens=self.cumulative_total_tokens,
+            )
+        )
+
+    def format_token_usage_log(self):
+        result = "step_name,"
+        result += "prompt_tokens_in_step,completion_tokens_in_step,total_tokens_in_step"
+        result += ",total_prompt_tokens,total_completion_tokens,total_tokens\n"
+        for L in self.token_usage_log:
```
AntonOsika, line 116: Prefer the lower case variable names a lot 😬

pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies = [
   'typer >= 0.3.2',
   'rudder-sdk-python == 2.0.2',
   'dataclasses-json == 0.5.7',
+  'tiktoken'
```
AntonOsika, line 22: Ah there should be a trailing comma here right?

gpt_engineer/ai.py
UmerHA: Haha, but ruff doesn't :D
I'll change the name to sth else

UmerHA: change to `log`

AntonOsika:
Fantastic

AntonOsika:
Merging — just checking, did you test this runs as expected, and if not could you do it? (Traveling for work atm)

UmerHA:
> Merging — just checking, did you test this runs as expected, and if not could you do it? (Traveling for work atm)

Of course I did :)

PR #444, opened 2023-06-29
Clearer tutorial instructions on readme

Jack-Lewis1:
For me, the instructions were unclear on how to run the example


JinchuLi2002 (changes requested):
Hi. This change assumes that users execute gpt-engineer from a cloned repo, and disregards the case when they chose to install the stable release via ```pip install gpt-engineer```. Maybe consider putting your change under 'development'?

AntonOsika:
if you pip install -e . the original is right

PR #445, opened 2023-06-29
Bug Fix: Model Name Should Not Be Verified Through OpenAI with Custom Endpoint

### LINKED ISSUE #331: adding a custom api end point
I use a custom and free api endpoint so I would like to add a feature  which is similar to:

```
export OPENAI_API_KEY=[your api key]
```

so I would like to have:

```
export OPENAI_API_BASE=[your custom api base url]
```

If you don't export it then it will have the default base url.

is it appropriate for me to work on this or if it is already implemented then please let me know.

JinchuLi2002:
Closes #331

## Overview
This Pull Request aims to fix the bug that prevents setting custom API endpoint.

## Background
Many open-source LLMs (e.g. FastChat, ChatGLM) provides OpenAI-compatible APIs, which means once these models are set up and running locally, users should be able to query them, still via ```openai.ChatCompletion.create()```,  simply by setting ```export OPENAI_API_BASE=<where model is running>```. However the current version of GPT-Engineer attempts to verify model name through OpenAI even if a custom api base is set.

## Changes 
* Modified such that model name is only verified (by ```fallback_model()```) only if a custom ```OPENAI_API_BASE``` is not set.

### DIFF
--- a/gpt_engineer/ai.py
+++ b/gpt_engineer/ai.py
@@ -4,13 +4,34 @@ import logging
 
 import openai
 
+import os
+
 logger = logging.getLogger(__name__)
 
+def fallback_model(model: str) -> str:
+    try:
+        openai.Model.retrieve(model)
+        return model
+    except openai.InvalidRequestError:
+        print(
+            f"Model {model} not available for provided API key. Reverting "
+            "to gpt-3.5-turbo. Sign up for the GPT-4 wait list here: "
+            "https://openai.com/waitlist/gpt-4-api\n"
+        )
+        return "gpt-3.5-turbo"
 
 class AI:
-    def __init__(self, model="gpt-4", temperature=0.1):
+    def __init__(
+            self,
+            model="gpt-4",
+            temperature=0.1,
+            openai_api_base=os.getenv("OPENAI_API_BASE", default=openai.api_base),
+        ):
         self.temperature = temperature
-        self.model = model
+
+        # model name is often irrelevant when querying locally deployed LLMs
+        self.model = fallback_model(model) if "api.openai.com/" in openai_api_base else model
+        openai.api_base = openai_api_base
 
     def start(self, system, user):
         messages = [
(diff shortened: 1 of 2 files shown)


AntonOsika:
Thanks @JinchuLi2002 !

The only problem with this is that the collect_learnings will not use the right model name if we move the fallback here.

To address, could for example use `ai.model`.

Question though: How do you propose we get the right model name when setting other OPENAI_API_BASE than the default? 

Would love to merge this if we can solve this!

JinchuLi2002:
@AntonOsika 
Hi Anton, thanks for the reply! :D

1. For your first point, could you be more specific how the issue could rise? The only impact with me moving ```fallback_model()``` with regards to ```collect_learnings()``` is that it could allow a non-OpenAI model name to be passed to the latter, if a custom end point is set. Do you mean that ```send_learning()``` would mandate the model name sent to be a OpenAI model?
2. Sadly, I don't think there's an effective way of verifying the model name with a custom endpoint. Because it differs with the local models which the user deploys. For example, the model name is 'codegen' for [FauxPilot](https://github.com/fauxpilot/fauxpilot/blob/main/documentation/client.md) and 'vicuna-7b-v1.3' for [Fastchat](https://github.com/lm-sys/FastChat/blob/main/docs/openai_api.md)

AntonOsika:
Should we maybe do whatever langchain does to standardize model selection?

JinchuLi2002:
@AntonOsika For LangChain I don't think they have a standardized model selection mechanism either. For openai models, it's defaulted to 'text-davinci-003' ([source](https://github.com/hwchase17/langchain/blob/master/langchain/llms/openai.py#L135)), and for google_palm model, it's straight up ```model_name: str = "models/text-bison-001"``` ([source](https://github.com/hwchase17/langchain/blob/master/langchain/llms/google_palm.py#L82)) and it's for the users to make sure the model_name is set correctly.

AntonOsika:
Hey there are a few things that would be needed to merge this:

1. Consider using OpenaiChat as a field of the AI class: https://github.com/hwchase17/langchain/blob/1f4a51cb9c84c6a061cda729c4cdd3783d98922c/langchain/llms/openai.py#L659
2. Pass in the base into AI, not get it from env variable


I think there is another PR that looks at this

JinchuLi2002:
thanks @AntonOsika !
will open a separate PR after I do more research

PR #446, opened 2023-06-29
remove a duplicate phrase from the prompt

MrF0o:
this phrase was written twice "Make sure that every detail of the architecture is, in the end, implemented as code."


AntonOsika:
Thanks! Ended up there when resolving conflicts.

PR #451, opened 2023-06-30
refactor(makefile): :recycle: remove repetitive command

k1lgor:
Removing repetitive command `dev-install` because it does the same as `install`. Does not matter if you run `make install` or `make dev-install`, both do the same thing.


kingh0730:
I believe having `dev-install` is for extensibility in the future.

k1lgor:
so can be added in the future

AntonOsika:
Great thanks for updating @k1lgor

PR #459, opened 2023-07-01
Added a new step `integration_testing` (self-healing)

pmdyy:
Based on `unit_tests` placed in the memory, this step will ask for a command (or snippet) necessary to run against it. 
Runs continuously until all test errors (or even compile errors) are fixed. This may replace the `fix_code` step found in `tdd+`.

### DIFF
--- a/gpt_engineer/steps.py
+++ b/gpt_engineer/steps.py
@@ -15,7 +15,8 @@ from gpt_engineer.learning import human_input
 
 def setup_sys_prompt(dbs: DBs) -> str:
     return (
-        dbs.preprompts["generate"] + "\nUseful to know:\n" + dbs.preprompts["philosophy"]
+        dbs.preprompts["generate"] + "\nUseful to know:\n" +
+        dbs.preprompts["philosophy"]
     )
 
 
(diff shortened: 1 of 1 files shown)


gpt_engineer/steps.py
```diff
@@ -15,7 +15,8 @@
 
 def setup_sys_prompt(dbs: DBs) -> str:
     return (
-        dbs.preprompts["generate"] + "\nUseful to know:\n" + dbs.preprompts["philosophy"]
+        dbs.preprompts["generate"] + "\nUseful to know:\n" +
+        dbs.preprompts["philosophy"]
```
AntonOsika, line 19: We're using line width = 90. It's set in pyproject.toml.

Do you think you could let your black respect this config? (otherwise the CI will fail on main, we don't want that)

AntonOsika (changes requested):

AntonOsika:
Great initiative @pmdyy. Lots of potential here!

It's not clear that this can help improve the code though. The prompt reads:

"\nPlease suggest ways to fix these issues by modifying `test.sh`."

While we want it to modify the code and/or the files.

Have you tried this?

Would be great to run benchmark.py on this (at least for a few cases) or see it in action over discord screenshare before we merge so we know it's working as expected!

AntonOsika:
Hey happy to hear your thoughts on the above

AntonOsika:
This has conflicts and seems stale for now, closing but feel free to reopen

PR #463, opened 2023-07-01
Add benchmark report

artmoskvin:
This PR addresses https://github.com/AntonOsika/gpt-engineer/issues/240 by adding the benchmark report after running the evaluation step. Additionally, this report can be inserted in the `benchmark/RESULTS.md` file if needed. The insertion is a bit hacky now because it is based on certain assumptions around the results file structure. I believe there's a better way of doing it so any feedback is appreciated. 

On a more general note, I was a bit confused by the difference between the `Works` and `Perfect` criteria. The current evaluation logic leaves `Works` empty if a user replies 'yes' to `Perfect` which is a bit misleading. I think if the code works it works and ideally we would measure that using tests which is probably part of the roadmap. Anyway, please take a look and let me know what y'all think 🙏

### DIFF
--- a/scripts/benchmark.py
+++ b/scripts/benchmark.py
@@ -1,13 +1,16 @@
 # list all folders in benchmark folder
 # for each folder, run the benchmark
 import contextlib
+import json
 import os
 import subprocess
 
+from datetime import datetime
 from itertools import islice
 from pathlib import Path
 from typing import Iterable, Union
 
+from tabulate import tabulate
 from typer import run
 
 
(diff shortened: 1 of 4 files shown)


scripts/benchmark.py
```diff
@@ -71,6 +74,75 @@ def main(
...
+    line_number = -1
+    for i, line in enumerate(lines):
+        if line.startswith(header_prefix):
+            line_number = i
+            break
+
+    if line_number != -1:
+        lines.insert(line_number, new_section)
+    else:
+        print(f"No section of level {level} found.")
```
AntonOsika, line 128:
```suggestion
        print(f"Markdown file was of unexpected format. No section of level {level} found. Did not write results.")
```

AntonOsika:
Thanks for the PR!

Suggested a small clarification to the print statement.

I agree with `works` defintion: It should be set to true if perfect is true.

Was unsure if I wanted to populate it if the user didn't explicitly set it, but now I think it should.

scripts/benchmark.py
MrDaVinC: u

captivus:
This series of comments smells spammy. @MrDaVinC please respond to this comment to substantiate your comments.

captivus:
> This series of comments smells spammy. @MrDaVinC please respond to this comment to substantiate your comments.

User blocked.

Issue #131, opened 2023-06-17
Can GPT engineer be used to modify or extend existing codebase instead?


morganavr:
Today I have come across this open source code completion tool - https://tzap.io/ which uses GPT API (3.5, 4)
They took a different approach of using embeddings and vector databases to scan existing codebase and send it along (or instead) with main prompt. Advantage of this approach as opposite to text-only prompts is that bigger amount of contextual information can be sent to OpenAI model. I was wondering if similar functionality can be implemented in `gpt-engineer` as well to extend its functionality from creating new projects to analyzing and modifying existing source code repositories.

jebarpg:
@morganavr https://github.com/AntonOsika/gpt-engineer/issues/79 https://github.com/AntonOsika/gpt-engineer/issues/119  https://github.com/AntonOsika/gpt-engineer/issues/95

There is an extension for VSCode : [Code GPT](https://codegpt.co/) which has this ability in part... you have to highlight a section of code for it to edit and/or refactor.

There is a way to get gpt-engineer to work on already established code base, but it takes some manual fiddling to do. The way they were re-running this issue's code: https://github.com/AntonOsika/gpt-engineer/issues/101#issuecomment-1595792433 would be similar to how you would run gpt-engineer over an existing code base. But there isn't an automatic process to do this yet. But it is doable. 

@patillacode @noxiouscardiumdimidium

morganavr:
> There is an extension for VSCode : [Code GPT](https://codegpt.co/) which has this ability in part... you have to highlight a section of code for it to edit and/or refactor.

The thing is - very often I need to write new functionality that is dependent on multiple other classes, structures, function calls and I can not simply highlight sections of the code, instead smart AI tool ideally should understand my requirements for new functionality and then analyze existing code base to understand how it can be implemented and integrated into existing code base.

@jebarpg

different55:
There's no AI in existence with enough context to understand any substantial codebase. If you set it free on an entire codebase it'd just make a horrific mess of it as huge swaths of it shimmered in and out of its short term memory.

zerofill:
> There's no AI in existence with enough context to understand any substantial codebase. If you set it free on an entire codebase it'd just make a horrific mess of it as huge swaths of it shimmered in and out of its short term memory.

I could see a possible way of it going through the existing code and creating basically bullet points to understand the codebase, and being more effective in creation things you would be adding though.

jebarpg:
[...]

Issue #95, opened 2023-06-17
Feature Request: read files from an existing workspace and request AI to make changes


NoCLin:
This will be very useful when we have made modifications to the workspace generated by AI.

We can then ask the AI to continue modifying the code, just like a PM.

psirdev:
I proposed it here: https://github.com/AntonOsika/gpt-engineer/issues/79

I think I can work on this, would be awesome

NoCLin:
> I proposed it here: #79
> 
> I think I can work on this, would be awesome

Thanks Pablo. Maybe we can do it together :)

clickbrain:
@psirdev - I posted this in #79, but wanted to post here for @NoCLin to see as well. 

I agree with this. I've used the standard ChatGPT interface to interact and create applications, but I almost always have to go back to the prompt with errors that resulted from running the code. Having a way to get ChatGPT to add functionality or fix errors would be great.

As an example, I just had ChatGPT create a Telegram Bot, but one of the libraries it uses is throwing an error when I execute main.py. When I looked it up, it looks like it is an error caused by capitalizing the word Filters. It would be great if I could start it up again and tell gptengineer, that I ran into this error and to fix it.

In addition, I know after using an application that I will want it to add new features or make changes to functionality and beig able to have it read in or remember what it did on a project would be ideal.

Finally, I would like to be able to tell it to provide deployment instructions. This is important, because some users of this are using it because they don't know how to code nor deploy an application to a server and in my specific case, deploy a bot in telegram, etc. I put those instructions in main and it added it as part of the specification, but did not provide any instructions.

clickbrain:
I've created an adjacent issue about getting ChatGPT to finish code #119

AntonOsika:
Solved 🚀

Issue #79, opened 2023-06-16
Analysis - Give context of a project to GPT Engineer


psirdev:
GPT Engineer is amazing. But right now the purpose is for small projects, projects where you need little implementations or requirements.

But... What about to give a full context of a project? If ChatGPT can understand what methods and classes has some projects on GitHub or packages in npm, maybe he can have a fully understand of a project and modify parts of it.

What about limits of ChatGPT prompt? We can give some prompts in several windows to give fully understanding of what's going on. 

I can work on this if someone has the courage to develop it with me.

clickbrain:
I agree with this. I've used the standard ChatGPT interface to interact and create applications, but I almost always have to go back to the prompt with errors that resulted from running the code. Having a way to get ChatGPT to add functionality or fix errors would be great. 

As an example, I just had ChatGPT create a Telegram Bot, but one of the libraries it uses is throwing an error when I execute main.py. When I looked it up, it looks like it is an error caused by capitalizing the word Filters. It would be great if I could start it up again and tell gptengineer, that I ran into this error and to fix it. 

In addition, I know after using an application that I will want it to add new features or make changes to functionality and beig able to have it read in or remember what it did on a project would be ideal. 

Finally, I would like to be able to tell it to provide deployment instructions. This is important, because some users of this are using it because they don't know how to code nor deploy an application to a server and in my specific case, deploy a bot in telegram, etc. I put those instructions in main and it added it as part of the specification, but did not provide any instructions.

jebarpg:
@psirdev @clickbrain https://github.com/AntonOsika/gpt-engineer/issues/101#issuecomment-1595792433
I believe it can be done with gpt-engineer. There is a key value pair database setup for keeping track of pass conversations and code generation which can be handing in implementing this. Also[ Code GPT](https://marketplace.visualstudio.com/items?itemName=DanielSanMedium.dscodegpt) has this sort of feature on selected sections of code in a file.

clickbrain:
Any idea how this can be done? 

I haven't tried this yet, but will give it a shot: Each step in steps.py will have its communication history with GPT4 stored in the logs folder, and can be rerun with scripts/rerun_edited_message_logs.py.

leomariga:
I put some consideration of this issue in #131 . 
I'm thinking of doing a proof of concept with some ideas mentioned in this discussion.
If anyone is developing something in this direction let me know so we can work together.

jebarpg:
[...]

PR #465, opened 2023-07-01
Improve existing codes

leomariga:
As requested in #79, #95 and #131 I did a proof of concept on how we could add new functionalities or fix errors from a set of existing code.

To use the implementation, we need to do the following setup:
 - In a file called `file_list.txt` in the root of your project, add a list of code file paths, one in each line. These files should be enough to give the AI agent context to the modification you want to do, but it does not need to be complete standalone execution. 
 - Using the `prompt` file we add a description of the modification we need to be done. It could be adding a docstring, adding new functions, classes, or even asking to correct a bug.
 - Call the modification process by using `gpt_engineer projects/test --steps improve_code`

In this stage of implementation, we are sending all code text in a single message which is not ideal, so it only works on code small enough to be inside a prompt message, but it is a start.

Here is a small example:

notepad_NI3fbpIOyP

Code_1eWl7Oh2oB

Code_pnyLvtvN7X

This PR is already a working solution. In a future modification, I plan to separate the codes into different messages to see the results.

What do you guys think?


k1lgor (commented):
A note for you, the naming convention for the methods is incorrect.

gpt_engineer/chat_to_files.py
```diff
@@ -40,3 +41,26 @@ def to_files(chat, workspace):
     files = parse_chat(chat)
     for file_name, file_content in files:
         workspace[file_name] = file_content
+
+
+# Get code content from a code list
+def getCodeStrings(input):
```
AntonOsika, line 47: get_code_strings

Python style

gpt_engineer/chat_to_files.py
```diff
@@ -40,3 +41,26 @@ def to_files(chat, workspace):
...
+    filesPaths = input["file_list.txt"].split("\n")
+    filesDict = {}
+    for filePath in filesPaths:
+        with open(filePath, "r") as file:
+            fileData = file.read()
+            fileName = os.path.basename(filePath).split("/")[-1]
```
AntonOsika, line 53: Can de-indent these for clarity

gpt_engineer/chat_to_files.py
```diff
@@ -40,3 +41,26 @@ def to_files(chat, workspace):
     files = parse_chat(chat)
     for file_name, file_content in files:
         workspace[file_name] = file_content
+
+
+# Get code content from a code list
+def getCodeStrings(input):
+    filesPaths = input["file_list.txt"].split("\n")
```
AntonOsika, line 48: This looks too cumbersome imo.

We should be able to generate this list of filepaths from the project directory to make it nice for the user!

Could, for example, print the content of the project directory, and ask the user 1. which folders to include and 2. which file extensions to include. And then this is stored to memory.

AntonOsika (changes requested):

AntonOsika:
Hey @leomariga great initiative!

Question:

How do you see this being used?

Basically, how does one set up gpt-engineer for a new project?

I'm thinking two options:
- create new project with prompt file + memory folder, where `workspace` symlinks to the existing codebase
- gpt-engineer creates a `.gpteng` folder that contains prompt and `memory`

WDYT? Maybe there is a third?

AntonOsika:
Also general feedback:
1. Python style convention is to use snake case
2. I think requiring a file_paths.txt is too complex for the user :)

AntonOsika:
Created a discussion in discord around this

leomariga:
Thank you for your feedback and review =) 

I agree with you that manually writing `file_paths.txt` is not the best user experience. I didn't put much thought into how would be the input method. I also liked the idea of using the `.gpteng` folder. 
 
(Brainstorming) We could do something like this:
 
1. At the root of the existing code repo we call something like `gpt-engineer -improve` which will create the `.gpteng` folder.
2. We write in the terminal what we want to modify. This string will be stored in the `prompt` file in `.gpteng` folder.
3. The tree of files of the project is displayed in the terminal and we select it with numbers (or we could even open the file explorer for the user!)
4. The gpt-engineer does its job and automatically replaces the file in the end.

I saw the #contributors channel in the discord, but I don't have access to it yet. Let's wait to see the ideas there. =)

gpt_engineer/chat_to_files.py
WIRENKARL: Yes thank you for stating the obvious.

leomariga:
Hey, thanks for all the feedback and ideas. I'm working on your suggestions.
 
I was thinking we could change the file `chat_to_files.py` to something like `utils.py` or `tools.py` and use it to put utility functions, like reading/parsing/writing files or strings. 

The methods I put there do not make sense with the file name. I think I should change the location of these methods or rename the file. What do you guys think?

I'm open to ideas.

AntonOsika:
I think what you say makes sense!

I really prefer specific file names though so it’s easy to find things. Could have chat_to_files and files_to_chat as separate files?

very happy you’re working on this, it’s the top requested feature by far.

AntonOsika:
As for which files to add, I think we should keep it simple at first and just:
- ask if all files in selected path argument should be included
- if not, let user use arrow keys and enter to toggle which files/folders should be included (I believe there must be a cli arrow select dependency we could use) 


Or we just do this:
We run the tree command, and then we open the output in a text editor, and then the user can delete rows form it and everything that is not deleted when editor is closed is used. 

What do you think?

gpt_engineer/chat_to_files.py
hrobbins: If only there were some gpt-engineer functionality that could do this for us. ;)

leomariga:
There you go. 
- Now we only need to call `gpt-engineer --improve` in the root folder of an existing project. A `.gpteng` folder will be created by gpt-engineer.
- I did some experiments in opening a selection window with python's tkinter instead of selecting the files with the terminal. I think it is much easier to use a gui, but we could implement both if someone insists on using the terminal.

I send a video of this feature working on the discord page.

hrobbins:
I've started on a version of this.  My approach is to start with the entry point(s), and iteratively include dependencies within the project.

leomariga:
> I've started on a version of this. My approach is to start with the entry point(s), and iteratively include dependencies within the project. Currently creates .filename.meta files for each analyzed file; not unlike .h library files you would have in C describing the contents of the associated .c file.

Nice. The file selection on this solution is the function `ask_for_files()`. We can add more input methods there if we want =)

gpt_engineer/main.py
```diff
@@ -36,6 +43,18 @@ def main(
     logging.basicConfig(level=logging.DEBUG if verbose else logging.INFO)
 
     input_path = Path(project_path).absolute()
+
+    # For the improve option take current project as path and add .gpteng folder
+    # By now, ignoring the 'project_path' argument
```
AntonOsika, line 48: Best way would be to always have it optional, but if it’s given them it should not be cwd for improve mode either I think!

gpt_engineer/chat_to_files.py
```diff
@@ -40,3 +43,83 @@ def to_files(chat, workspace):
...
+            )
+            dbs.workspace[file_name] = file_content
+
+
+def get_code_strings(input) -> dict[str, str]:
+    """
+    Read file_list.txt and return file names and its content.
+    """
+    filesPaths = input["file_list.txt"].strip().split("\n")
+    filesDict = {}
```
AntonOsika, line 78: Python style is snake case 😬

AntonOsika: A few more occurances if this

gpt_engineer/chat_to_files.py
```diff
@@ -40,3 +43,83 @@ def to_files(chat, workspace):
...
+        else:
+            # If the AI create a new file I don't know where to put it yet
+            # maybe we can think in a smarter solution for this in the future
+            # like asking the AI where to put it.
+            #
+            # by now, just add this to the workspace inside .gpteng folder
+            print(
+                f"Could not find file path for '{file_name}', creating file in workspace"
+            )
+            dbs.workspace[file_name] = file_content
```
AntonOsika, line 70: Hm I think workspace should be == cwd no?

AntonOsika: I’d be fine with not checking and just using chat_to_files

the improvement might be to refactor into new files right?

gpt_engineer/chat_to_files.py
```diff
@@ -40,3 +43,83 @@ def to_files(chat, workspace):
...
+    """
+    return filestr
+
+
+def ask_for_files(input) -> dict[str, str]:
+    """
+    Display a tkinter file selection window to select context files.
+    Return a dictionary of file_name as key and file_path as value
+    """
+    root = tk.Tk()
```
AntonOsika, line 107: I would prefer if we can run GPT engineer via ssh, ie no GUI. To facilitate that I’d suggest we ask user if they:
1. want to launch a gui selector
2. Want to reuse the file that contains files to use, or fallback to all files if that file does not exist

AntonOsika: Also nice to not launch the selector every time

AntonOsika (changes requested):
Nice work!

I left a few comments. Not all are necessary to fix to merge.

what I would also suggest to make it cleaner is this:

workspace is == cwd when improve is set. And we use chat_to_files like normal when we run improve.

AntonOsika:
Left some comments! 

Also adding some thought:

would be great to automatically print the diffs from the generation and ask the human if they like the diffs or not

leomariga:
Quick progress update. Now we can choose how to select the files as follows:

mspaint_W5Z1xZnYoP

Still working on other stuff.

lectair:
I want to collaborate on this, please upload PR when available.

lectair:
> A note for you, the naming convention for the methods is incorrect.

If that's important enough, maybe you'd want to fix it.

lectair:
I am also working on this feature. I have sent you a pull request, @leomariga. @AntonOsika, do you think it would be a good idea to create a branch for this?

lectair:
I think creating a CLI that you can interact with while gpt-4 is editing the code to be able to iterate indefinitely and tell it to delete, add or change anything. This could make the process of developing software really efficient. Something like a ChatGPT but with code. Like the code interpreter of ChatGPT Plus but better :)

But to achieve this it would be important to get it to always run the code correctly, without problems finding paths, setting up files or doing extra commands.

AntonOsika:
Preparing review...

AntonOsika:
## PR Analysis

- 🎯 **Main theme:** Adding functionality to improve existing code
- 🔍 **Description and title:** Yes
- 📌 **Type of PR:** Enhancement
- 🧪 **Relevant tests added:** No
- ✨ **Focused PR:** Yes, the PR is focused on adding a new feature to improve existing code. All changes are related to this feature.
- 🔒 **Security concerns:** No, the PR does not introduce any obvious security concerns. However, it's always a good practice to handle file operations carefully to prevent any potential security issues.
## PR Feedback

- 💡 **General PR suggestions:** The PR is well-structured and the code changes are well-documented. However, it lacks tests to ensure the new functionality works as expected. It would be beneficial to add unit tests for the new functions and integration tests to ensure the new feature works with the existing code.

- 🤖 **Code suggestions:**


  - **relevant file:** gpt_engineer/chat_to_files.py
   **suggestion content:** Consider handling exceptions when opening and writing to files. This can prevent the program from crashing if there are issues with file permissions or if the file does not exist. [important]


  - **relevant file:** gpt_engineer/file_selector.py
   **suggestion content:** The `ask_for_files` function could be refactored to reduce its complexity. Consider breaking it down into smaller, more manageable functions. This would improve readability and maintainability. [medium]


  - **relevant file:** gpt_engineer/main.py
   **suggestion content:** Consider adding a validation for the `improve_option` argument. If it is not a boolean, the program may behave unexpectedly. [medium]


  - **relevant file:** gpt_engineer/steps.py
   **suggestion content:** The `improve_existing_code` function is quite long and does a lot of things. Consider breaking it down into smaller functions to improve readability and maintainability. [medium]

AntonOsika:
Great to hear, screenshot looks good!

AntonOsika:
Hey @leomariga – any news on this?

gpt_engineer/chat_to_files.py
leomariga: The problem I see is the creation of "all_output.txt" in the root of the folder. 

Also, if I only use "to_files" function I'll be limited to the files from the workspace. With the function "ovewrite_files" we can change files in any folder. 

By now I'll use the workspace as cwd, but I'll continue using the overwrite function.

gpt_engineer/main.py
leomariga: By default, we use the "example" folder if nothing is passed in file_path. So we have two options:

1- If we receive the "example" folder as file_path, we assume there is no argument in the function call. In this case, we assume the user is using cwd as workspace. If something is different from "example" as input, we use that file path as workspace. The consequence of this option is that we will never be able to improve the "example" folder because it will be invisible to the script. 

2- We remove the "example" folder as the default file path. In this case, if the user uses an argument, we use it as workspace. If no argument is being set, we use cwd as workspace.

leomariga:
Done. 

I just want to point out that @lectair did interesting modifications in a PR to my branch, but I didn't get a response from him after a review I needed to merge into my branch. 

I think we can open a new PR with his changes in the official repo when this PR is merged.

leomariga:
Merged with commits from @lectair =)

gpt_engineer/chat_to_files.py
AntonOsika: We can just skip writing all_output.txt. It should be in memory/logs regardless!

AntonOsika: Would be so nice to simplify this function!

Less code is always better.

I.e.: We set workspace to the current directory, and get rid of the "open(filename)" boilerplate by always doing `dbs.workspace[name] = content`

gpt_engineer/chat_to_files.py
```diff
@@ -40,3 +41,57 @@ def to_files(chat, workspace):
     files = parse_chat(chat)
     for file_name, file_content in files:
         workspace[file_name] = file_content
+
+
+def overwrite_files(chat, dbs, replace_files):
+    """
+    Replace the AI files to the older local files.
+    """
+    dbs.workspace["all_output.txt"] = chat
```
AntonOsika, line 50: Can simply stop doing this.

AntonOsika:
Hey Leo great to see this ready to be merged.

Last step, which will take some time, is to merge / rebase in main and resolve conflicts so we can merge.

While doing it I have some quick final improvements:

- Stop using a directory called workspace when we run the -i command: It should always be run from the directory that one is currently in. Currently, a folder called workspace is created.
- See my comments in the PR

Also, small optional improvements:
- Make the "user file_txt" option for selecting files first and default
- Number the file selection alternatives from 1 (not 0)
- Consider not listing any `venv` or `node_modules` folders (as they have so many files)

AntonOsika:
One more thing:

I think we should simplify the "improve code" prompt, and use another final system message focused on explicitly requesting the output to be in the format we want, just like the `gen_code` step does (it uses the `use_qa` prompt).

When I tried it, it didn't give the right format (see screenshot)
image

Let's go 🏃

AntonOsika:
Marking this as stale. 

Would be great to pick it up, hopefully we don't have to close it.

PR #475, opened 2023-07-03
Main

Nick777-Pixel:


local.env
esillen, line 1: ARGH!! Don't commit this!!!

AntonOsika: 🙈 

Please revoke your API key @Nick777-Pixel

AntonOsika: https://openai.com/waitlist/gpt-4-api

AntonOsika:
already on main

PR #499, opened 2023-07-05
.env setup for OpenAI API key

OriAshkenazi:
This pull request adds support for loading environment variables from a .env file in the Makefile of the gpt-engineer project. This is achieved by using the python-dotenv package to read environment variables from a .env file when the make run command is executed. The following changes have been implemented:

1. Updated the Makefile: Added a new load-env target in the Makefile. This target uses python-dotenv to load the environment variables from a .env file.
2. Updated the run target: Modified the run target to depend on the load-env target. This ensures that the environment variables are loaded before the run command is executed.
3. Updated pyproject.toml: Added python-dotenv >= 1.0.0 to the list of project dependencies in pyproject.toml. This ensures that python-dotenv is installed when setting up the project.
4. Updated README.md: Updated the new setup process to reflect the changes.

These changes allow for better management and confidentiality of environment variables in the gpt-engineer project, particularly when it comes to sensitive information that should not be hard-coded. This method will also help users keep track of their API key between runs.


AntonOsika:
Thanks @OriAshkenazi !

I like the idea I’m not sure it covers all ways to run it though. 

It should be possible to pip install and then only run gpt-engineer. No make. 

I thought about this last night though and what I think we want is:

 1. Check if API KEY env variable exist
 2. If it doesn’t, call load_dotenv
 3. Package these two steps in a nice function so it is used in main.py and all scripts etc.

AntonOsika:
What do you think?

OriAshkenazi:
> What do you think?

I'll give it another go on Sunday to make sure it covers the scenario you mentioned! :)

OriAshkenazi:
Updated
What do you think? I feel like everything is covered now

Makefile
```diff
@@ -34,29 +34,35 @@ create-venv:
...
+install-pre-commit: load-env
 	@echo -e "$(COLOR_CYAN)Installing pre-commit hooks...$(COLOR_RESET)" && \
 	source venv/bin/activate && \
 	pre-commit install
 
+#Defines a target named load-env. This target will load the environment variables from the .env file.
+load-env:
+	@echo -e "$(COLOR_CYAN)Loading environment variables...$(COLOR_RESET)" && \
+	source venv/bin/activate && \
+	python -m dotenv load
```
AntonOsika, line 58: Do we still need this? Would be great to keep this file as simple as possible.

README.md
```diff
@@ -37,15 +37,13 @@ For **development**:
 
 **Setup**
 
-With an OpenAI API key (preferably with GPT-4 access) run:
+- Duplicate the `.env.template` file and rename it to `.env`.
+- Open the `.env` file and replace `$key` with your OpenAI API key (preferably with `gpt-4` access). The key should be set as follows: `OPENAI_API_KEY=your_key_here`. Make sure there are no spaces around the equals sign and the key is enclosed in quotation marks if it contains special characters.
+  - If you want to set the OpenAI API key as an environment variable directly, you can do so using the following command in your terminal: `export OPENAI_API_KEY=your_key_here`. Please note that this environment variable will be available only for the current session. If you close the terminal or start a new session, you will need to set it again.
+  - `gpt-engineer` will first check if the `OPENAI_API_KEY` environment variable is set. If it is not, it will load the API key from the `.env` file.
+  - Note that `gpt-3.5-turbo-16k` is set as a fallback model for API keys without `gpt-4` privileges.
```
AntonOsika, line 44: We've tried hard to make the README.md as short as possible so that people take time to read everything!

Hence, I'm thinking, could we simplify this?
And/or move some of it to a separate docs/CONFIG_README.md.

gpt_engineer/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    if os.getenv("OPENAI_API_KEY") is None:
+        load_dotenv()
+
+        # After attempting to load from .env, check again
+        if os.getenv("OPENAI_API_KEY") is None:
+            raise ValueError("Cannot run the program without OPENAI_API_KEY environment variable.\nPlease set it in your environment or in a .env file.")
+
+# Call the function when package is imported
+check_env_vars()
```
AntonOsika, line 17: We have this code duplicated. According to DRY we should only have it in one place!

It's not obvious, but I thought about it and I think this is the behavior we want:
- We move the function in main.py to its own file
- We import that and run it in the main() function and in the scripts that need it
- We remove the code in __init__

In this way things are as explicit not implicit according to python philosophy.

gpt_engineer/main.py
```diff
@@ -26,6 +41,9 @@ def main(
 ):
     logging.basicConfig(level=logging.DEBUG if verbose else logging.INFO)
 
+    # Call the new function before any actions that require environment variables
```
AntonOsika, line 44: Coding convention is to only add comments for "complex" pieces of code – this function name is so good it's self explanatory

```suggestion
```

AntonOsika:
Great. Appreciate the effort @OriAshkenazi.

Be prepared it often takes a few iterations to make things work exactly as we want them in serious development.

I added some comments since a part of the philosophy is to minimize the LoC.

AntonOsika (changes requested):

AntonOsika:
PS:

pytest failing due to that env variable doesn't exist for tests

OriAshkenazi:
> PS:
> 
> pytest failing due to that env variable doesn't exist for tests

How can I solve this without giving away an API key to the public? Can you explain to me what can be done? Even a general broad explanation will be good.
I'll make sure to attend to the rest of the comments later today.

AntonOsika:
Hey! I explained the proposed solution in my proposal in the review comments, did you see?

PR #512, opened 2023-07-07
Langchain integration

UmerHA:
Strongly based on https://github.com/AntonOsika/gpt-engineer/pull/466 - **shoutout to @hrobbins!**

---
**What it does:**
Enables using any model that's available in LangChain as backend.

**How to use it:**
- For gpt4/3.5 it's the same usage, i.e. `gpt-engineer <path_to_project> gpt-4`
- For other models, you have to provide a `<model_name>.yaml` file with the model configs. Then `gpt-engineer <path_to_project> <model_name> --modeldir <path_to_model_config_file>`

**How it works:**
- gpt-engineer search for a <model_name>.yaml and passes its content to LangChain, which then creates the model.
- per default the <model_name>.yaml is searched for in the package folder /models.
- `ai.py` has been adapted to use LangChain

In the future, more model configs can be added, so more models are available by default.


UmerHA:
need to fix typing issues

gpt_engineer/ai.py
```diff
@@ -23,72 +39,134 @@ class TokenUsage:
...
+
+    @staticmethod
+    def serialize_messages(messages: List[Message]) -> str:
+        return json.dumps(messages_to_dict(messages))
+
+    @staticmethod
+    def deserialize_messages(jsondictstr: str) -> List[Message]:
+        return AI.parse_langchain_basemessages(
+            messages_from_dict(json.loads(jsondictstr))
+        )
```
AntonOsika, line 143: Can we just do:

`return list(messages_from_dict(json.loads(jsondictstr))` ?

AntonOsika: And delete below function

gpt_engineer/ai.py
```diff
@@ -23,72 +39,134 @@ class TokenUsage:
...
+        parsed_messages = []
+
+        msg_type_to_cls = {
+            msg_cls(content="").type: msg_cls
+            for msg_cls in [AIMessage, HumanMessage, SystemMessage]
+        }
+        msg_types = list(msg_type_to_cls.keys())
+
+        for m in messages:
+            if m.type not in msg_types:
```
AntonOsika, line 158: Don't see this being necessary. We can assume it will work.

gpt_engineer/ai.py
```diff
@@ -152,3 +227,14 @@ def fallback_model(model: str) -> str:
...
+
+def serialize_messages(messages: List[Message]) -> str:
+    return AI.serialize_messages(messages)
+
+
+def cleanup_reponse(response: str) -> str:
+    response = re.sub(
+        "\\n", "\n", response
+    )  # for some reason models sometimes return \n instead of newline?
+    return response
```
AntonOsika, line 240: This will break when there is code that contains strings with \n no?

Would prefer if we let the LLM object handle this

models/gpt-4.yaml
```diff
@@ -0,0 +1,11 @@
...
+#best_of: 1
+frequency_penalty: 0.0
+#max_tokens: 8192
+model_name: gpt-4
+n: 1
+presence_penalty: 0.0
+request_timeout: null
+temperature: 0.0
+top_p: 1.0
+streaming: True
```
AntonOsika, line 11: Could you delete these files

AntonOsika (changes requested):

gpt_engineer/ai.py
```diff
@@ -23,72 +39,134 @@ class TokenUsage:
...
 
-    def next(self, messages: List[Dict[str, str]], prompt=None, *, step_name=None):
+    def combine_messages(self, messages: List[Message]) -> str:
+        msg_dict = messages_to_dict(messages)
+        logging.debug(msg_dict)
+        prompt = "\n".join(
+            "%s: %s" % (md["type"], md["data"]["content"]) for md in msg_dict
+        )
+        logging.debug("Prompt: " + prompt)
+        return prompt
```
AntonOsika, line 102: I'd prefer we get rid of this and only support GPT4 for now. It's by far the best code generation model.

gpt_engineer/ai.py
```diff
@@ -23,72 +39,134 @@ class TokenUsage:
...
+            messages=messages, answer=response, step_name=step_name
         )
 
         return messages
 
-    def update_token_usage_log(self, messages, answer, step_name):
+    def last_message_content(self, messages: List[Message]) -> str:
+        m = messages[-1].content.strip()
+        logging.info(m)
+        return m
```
AntonOsika, line 133: I think this function makes the code more difficult to read for a human.

Would prefer to remove it!

And `logging.info(m)` will be confusing no?

AntonOsika:
Good job @UmerHA! Soon there and merged 🏃 

The big concern from my side is the "model_dir" and "model_name" and that gpt4 is not supported.

Actualy, it's not on our roadmap to add open source models.

Also, I'd like to keep the code as simple as possible, and I'm not sure it is worth it for anyone to use local models? GPT4 is much better.

I'd propose we remove that part for now and potentially bring the local model stuff in a separate PR.

Thoughts?

hrobbins:
Should be a model for gpt4.

hrobbins:
My general perspective, I'm not sure what disadvantage there would be in
including it.  I can think of five contributors who would like to see it
included to facilitate their current efforts.  From what I understand, this
functionality was a key motivation for @UmerHA <https://github.com/UmerHA> to
do the work he has done.  By including it, you will likely accelerate GPT
Engineer development.  I personally have been talking to a few companies
that would like to pilot something like GPT-Engineer and possibly subsidize
it's development, as long as they can be confident their codebase isn't
going to end up being used as the next generation of training data (i.e.
not using third-party APIs).  Facilitating people testing alternative
models will also likely benefit GPT Engineer in the long run as the models
inevitably become more accessible on more common hardware.  And more
companies feeling comfortable they can control where their codebase is
going, the more corporate resources will be put behind the project.

UmerHA:
@AntonOsika Taking a step back: What's the reason for integrating with LangChain if we don't want other models?
My understanding was that's why we're doing https://github.com/AntonOsika/gpt-engineer/pull/466, on which this PR is based.

Apart from that I agree that gpt4 is currently the best model, but I would expect that non-openai models (incl. open-source) get more capable over time, so I would vote to enable them.

Also, is there a roadmap I can consult? Of course don't want to build stuff not on the roadmap. :)

AntonOsika:
Hey cool didn’t know there was a gpt4 model accessible in this way? Does it work out of the box with this approach. 

As long as it didn’t interfere with current way of using its great, happy to merge after some other review comments are addressed. it’s just that the way of putting messages together seems like it shouldn’t be supported by gpt4?

also — thanks both for contributions on this

hrobbins:
Worked last time I checked, but doesn't appear to work at the moment with
any branches of gpt engineer code base. Is it just me or did OpenAI quietly
roll back access to gpt-4 via the API?

jenspacito (approved):

gpt_engineer/ai.py
UmerHA: Unfortunately no. Doing `list(...)` gives a `List[BaseMessage]`, but we need a `List[Message]` (which is defined as `List[Union[AIMessage, HumanMessage, SystemMessage]]`).

The complexity comes from the fact that `List` is [invariant](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)), meaning `List[Some_Subclass]` is not accepted where `List[Some_Class]` is expected.

For context:
- `BaseMessage` is a LangChain class and the superclass of the LC classes `AIMessage`, `HumanMessage`, and `SystemMessage`
- `Message` is defined in this PR and is an alias for `Union[AIMessage, HumanMessage, SystemMessage]`

We have to choose between one of these options:
1. Don't do typing at all
2. Use same type as LC: Use `Sequence[BaseMessage]` in gpt-engineer. This works because `Sequence` unlike `List` is covariant. But Sequence is immutable. So everywhere we add to the messages, we would have to convert to `List`, do the operation, then re-convert to `Sequence`. Imo this makes the code complex and repetitive.
3. Parse at the point of interaction with LC: Parse between LC's `Sequence[BaseMessage]` and our `List[Message]` only at the LC interface. Then we can keep our code simple.

I think 3 is cleanest, so I chose that.

What are your thoughts?

gpt_engineer/ai.py
UmerHA: Agree & changed

gpt_engineer/ai.py
UmerHA: 1. Agree & deleted `logging.info(m)`
2. Disagree with `last_message_content` making the code more difficult to read. We're using that function 11 times (see image), so think makes sense to not always rewrite it, and the function name is pretty self-explaining IMO. But will do what you decide. Lmk!

image

gpt_engineer/ai.py
UmerHA: Agreed & changed

models/gpt-4.yaml
UmerHA: Done

gpt_engineer/ai.py
UmerHA: Agreed & deleted

hrobbins:
Another (recommended) alternative would be to use a data model like
ChatMessageHistory, already provided by Langchain, or roll your own.

-H

gpt_engineer/ai.py
```diff
@@ -23,72 +39,134 @@ class TokenUsage:
...
+                f"Unable to load LLM {model_name} from file {llm_filename}", e
+            )
 
         # initialize token usage log
         self.cumulative_prompt_tokens = 0
         self.cumulative_completion_tokens = 0
         self.cumulative_total_tokens = 0
         self.token_usage_log = []
 
+        # ToDo: Adapt to arbitrary model, currently assumes model using tiktoken
```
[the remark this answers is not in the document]
UmerHA: Unfortunately, LangChain doesn't give access to its models' tokenizers. So we would either (i) wait until LC does so, or (ii) implement each tokenizer for each model into gpt-engineer. Imo (ii) doesn't make sense because we want to let LC handle all the interactions with models, and focus on coding-specific stuff.

UmerHA:
Hey @AntonOsika, I incorporated your feedback.

I also removed the `model_dir` & `gpt4/3.5.yaml` files, because LangChain doesn't offer loading from config files for ChatModels yet.

gpt_engineer/ai.py
AntonOsika: Right the typing. My thinking is that we use typing.cast() (and assert if necessary)?

Otherwise the code is very confusing to the reader and I'd even prefer doing 1!
(which is a legit option: just adding  # type: ignore

gpt_engineer/ai.py
AntonOsika: Great on 1.

On the other, it's just that it's:

1. More characters to type
2. Needs an external dependency (AI)
3. which is an object? (even though it doesn't need to be initialized)
4. hides away unnecessarily, how simple everything actually is: It's just a list of messages! (knowing this makes it easier to understand the code)

gpt_engineer/learning.py
```diff
@@ -153,21 +153,14 @@ def ask_if_can_store() -> bool:
...
-        chunks.append(format_messages(messages))
+        chunks.append(logs[step.__name__])
     return "\n".join(chunks)
 
 
-def format_messages(messages: List[dict]) -> str:
-    return "\n".join(
-        [f"{message['role']}:\n\n{message['content']}" for message in messages]
-    )
-
```
AntonOsika, line 169: This breakes tests. See CI.

Any reason of changing it?

AntonOsika:
Cool! 🏃 

To be honest I think AI abstraction is a bit too complex right now and would like a proposal of a refactor. But I think it's OK we merge after we address current comments.

As @hrobbins points out I think ChatMessageHistory would be great.

gpt_engineer/learning.py
UmerHA: Deleting `format_messages` is not the issue. That function is not used anywhere. That's why I deleted it.

The test `test_collect_learnings` breaks because in the test we're doing sth like
```
code = "this is output\n\nit contains code"
code_with_extra_stuff = "extra stuff" + json.dumps(code)
assert code in code_with_extra_stuff
```
but `json.dumps` escapes the `\n` and turns it into `\\n`. That's one of the tests that fail which iiuc don't relate to the changes in this PR.

I will fix the tests by changing `assert code in code_with_extra_stuff` to `assert json.dumps(code) in code_with_extra_stuff`

gpt_engineer/ai.py
UmerHA: Okay, get your point, will change!

AntonOsika:
Let me know when you want me to take a look! 🙏

gpt_engineer/steps.py
```diff
@@ -293,7 +297,6 @@ class Config(str, Enum):
         gen_clarified_code,
         gen_entrypoint,
         execute_entrypoint,
```
AntonOsika, line 299:
```suggestion
        execute_entrypoint,
        human_review,
```

AntonOsika:
Happy birthday @UmerHA
And great work! 🚀

Same with me – very happy to get on top of PRs and see this.

Let's merge 🏃

PR #529, opened 2023-07-13
Add docker support.

ayoubfletcher:
I have created a Dockerfile to build the project into a docker image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,6 @@ With an OpenAI API key (preferably with GPT-4 access) run:
 - `export OPENAI_API_KEY=[your api key]`
 
 Alternative for Windows
-- `set OPENAI_API_KEY=[your api key]` on cmd
-- `$env:OPENAI_API_KEY="[your api key]"` on powershell
 
 **Run**:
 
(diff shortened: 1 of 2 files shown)


shubham-attri (approved):

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.11.4-alpine3.18
+
+WORKDIR /app
+
```
karlderkaefer, line 4: bash required when running example 
```
Executing the code...

Note: If it does not work as expected, consider running the code in another way than above.

You can press ctrl+c *once* to stop the execution.

/bin/sh: bash: not found
```

README.md
```diff
@@ -64,6 +64,17 @@ To get started, create a codespace for this repository by clicking this 👇
...
 [![Open in GitHub Codespaces](https://github.com/codespaces/badge.svg)](https://github.com/AntonOsika/gpt-engineer/codespaces)
 
+## Getting Started with Docker
+
+**Building the image**:
+- `git clone https://github.com/AntonOsika/gpt-engineer.git`
+- `docker build --rm -t gpt-engineer .`
+
+**Running the image**:
+- `docker run -it --rm -e OPENAI_API_KEY="YOUR OPENAI KEY" -v ./your-project:/project gpt-engineer`
```
karlderkaefer, line 74: the files that are created from container within volume are written as root user
```
drwxr-xr-x 4 root      root      4096 Jul 16 13:58 workspace
```
this make it's harder to modify, delete or execute files after creating. I would suggest to use this on Linux
```suggestion
- `docker run -it --rm -e OPENAI_API_KEY="YOUR OPENAI KEY" -u $(id -u):$(id -g) -v ./your-project:/project gpt-engineer`
```

karlderkaefer: and when creating own project the files must be open for any user and group so that root can write
```
chmod a+rw -R my-project
```

AntonOsika (changes requested):
See proposed comments by @karlderkaefer 👌

PR #537, opened 2023-07-16
Update README.md

apurvjha123:
There was missing windows script . It will be helpful for Window user .


AntonOsika:
Thanks for noticing, fixed with a link instead. Feel free to add to windows_readme.md

apurvjha123:
Everything seems fine but WINDOWS_README is not linked in README . So, I generated a pull request . Look into it .

PR #539, opened 2023-07-17
Add support for in-memory `DB`s

jesse-michael-han:
This allows proposed file changes to be buffered in memory and makes `gpt-engineer` compatible with other tools that buffer / stage proposed code changes, such as language servers.

Used by the `gpt-engineer` integration into the Rift language server and VSCode extension: https://github.com/morph-labs/rift/pull/71

### DIFF
--- a/gpt_engineer/__init__.py
+++ b/gpt_engineer/__init__.py
@@ -1,1 +1,1 @@
-__author__ == "morph"
+__author__ = "morph"
--- a/gpt_engineer/__init__.py
+++ b/gpt_engineer/__init__.py
@@ -1,0 +1,1 @@
+__author__ == "morph"
--- a/gpt_engineer/db.py
+++ b/gpt_engineer/db.py
@@ -3,23 +3,26 @@ import shutil
 
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Any, Dict, Optional
 
 
 # This class represents a simple database that stores its data as files in a directory.
 class DB:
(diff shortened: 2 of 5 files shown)


jesse-michael-han:
The addition to the prompt makes it so that `gpt-3.5-turbo` can generate files in the format required by `parse_chat` much more reliably.

AntonOsika:
Hey @jesse-michael-han !

The purpose wasn't clear when I first looked at it hence slow reply.

Quick questions:

Why don't you just replace the usage of `DB` class with a normal dictionary, or a thin wrapper around a normal dict, in the case of wanting buffered reads / writes?

Could update the typehints to allow for this!

jesse-michael-han:
Hey @AntonOsika, the in-memory dict was the quickest / most minimal code change to get the integration working.

It might be better to do it by generalizing the `DB` class to an `AbstractDB` class and having `FilesystemDB` and `InMemoryDB` instances of `AbstractDB`, with the choice of which one to create determined by a command line flag. 

The state of the branch reflects a bunch of quick changes we made to ship gpt-engineer with Rift 2.0. I'll make a new PR so that hopefully soon Rift can just use base `gpt-engineer`.

AntonOsika:
Looking forward to PR that addresses @jesse-michael-han!

Closing for aesthetics.

BTW:

Feel free to let DB be in memory version that inherits from dict, and DiskDB like below

```
class DB(dict):
    pass
    
class DiskDB(DB):
   def __init__(self):
       # override methods to current version of DB here....
```

PR #543, opened 2023-07-17
added support for C/C++ Makefile

mdabir1203:
When compiling C codes. gpt-engineer doesn't make the Makefile


mdabir1203:
?

AntonOsika:
Looks great.

The path forward here is to allow to inject one's one philosophy. e.g. you should easily have a config that includes Makefile preference.

Scope of current philosophy is js/ts/python just so that we don't bloat it with all languages

PR #555, opened 2023-07-25
codeql.yml

Profesor09:
add optimal timing for Ryan Fullen


Profesor09 (commented):
good to go

AntonOsika:
We don't have any java or JS.

Would you be able to explain the impact of the moved `)` and if so reopen?

PR #556, opened 2023-07-27
Minor fixes to documentation for Windows users

Oxymoron290:
specified which command must be ran as an administrator and fixed a broken link in the documentation.


AntonOsika:
Great thanks @Oxymoron290

PR #578, opened 2023-08-09
Rebase of @leomariga's Existing Project 

pbharrin:
As requested by @AntonOsika I've rebased @leomariga 's "Improve Existing Codes" PR #465 to recent changes.  

I tested and it works in "normal" mode (generate new code) and it works in Improve Existing Code mode.  Existing code mode works for command line mode as well as the TKInter based GUI.  I ran tests and they all passed (except test_ai.py which has nothing in it.)

I think there are some existing comments from the PR which could be implemented.  I will check on those.


pbharrin:
I think I have addressed everything in the last comments of #465.  

Additionally if the file names are not known or a new file is generated then these new files are placed in `workspace`.  This is not optimal, I know but we should discuss a solution.  The code does not know what is the correct place to place these files.  We could perhaps ask the AI or we could try to infer a root directory based on the largest common path from all the files, or some combination of these two.

leomariga:
Thanks for helping us with this issue @pbharrin . I hope we can see it merged soon. =)

AntonOsika:
Awesome great job @pbharrin 

I'll try to do the review later today, day full of afk duties.

gpt_engineer/chat_to_files.py
```diff
@@ -40,3 +41,57 @@ def to_files(chat, workspace):
...
+
+
+def get_code_strings(input) -> dict[str, str]:
+    """
+    Read file_list.txt and return file names and its content.
+    """
+    files_paths = input["file_list.txt"].strip().split("\n")
+    files_dict = {}
+    for file_path in files_paths:
+        file_data = ""
```
AntonOsika, line 78:
```suggestion
```

gpt_engineer/main.py
```diff
@@ -21,11 +23,34 @@ def main(
...
+    memory_path = input_path / "memory"
+    workspace_path = input_path / "workspace"
+    # For the improve option take current project as path and add .gpteng folder
+    # By now, ignoring the 'project_path' argument
+    if improve_option:
+        input_path = Path(os.getcwd()).absolute() / ".gpteng"
+        input_path.mkdir(parents=True, exist_ok=True)
+        # The default option for the --improve is the IMPROVE_CODE, not DEFAULT
+        # I know this looks ugly, not sure if it is the best way to do that...
+        # we can change that in the future.
```
AntonOsika, line 48:
```suggestion
```

gpt_engineer/main.py
```diff
@@ -21,11 +23,34 @@ def main(
...
+        if steps_config == StepsConfig.DEFAULT:
+            steps_config = StepsConfig.IMPROVE_CODE
+        memory_path = input_path / "memory"
+        workspace_path = Path(os.getcwd()).absolute()
```
AntonOsika, line 52: This will be overwritten by line 59-61

Could you put line 59-61 in an else clause?

gpt_engineer/preprompts/implement_on_existing
```diff
@@ -0,0 +1,19 @@
...
+```
+CODE
+```
+
+After that, implement the requested functionality by using the received code as base.
+The received code not necessarily will be a standalone implementation, it could be part of a bigger project which you might not have all files. Trust that everything you need to know inside the received files.
+Follow the same coding language and framework observed in the received code, using appropriate best practice file naming convention.
+Make sure that code in different files are compatible with each other.
+Ensure to implement the requested modification, if you are unsure, write a plausible implementation.
+Before you finish, double check that all parts of the architecture is present in the files.
```
AntonOsika, line 19: I think we def. need to experiment to get this prompt right.

Have you tried it a few times?

I would consider being inspired by the prompt that aider or mentat uses (google and check their code)

projects/example/prompt
```diff
@@ -1 +1 @@
-We are writing snake in python. MVC components split in separate files. Keyboard control.
+The current game boundry is 30x20 in main.py, please make that 40x40.
```
AntonOsika, line 1:
```suggestion
We are writing snake in python. MVC components split in separate files. Keyboard control.
```

AntonOsika: Let's fix this later

gpt_engineer/main.py
```diff
@@ -21,11 +23,34 @@ def main(
...
+        "-i",
+        help="Improve code from existing project.",
+    ),
     verbose: bool = typer.Option(False, "--verbose", "-v"),
 ):
     logging.basicConfig(level=logging.DEBUG if verbose else logging.INFO)
 
-    model = fallback_model(model)
+
+    input_path = Path(project_path).absolute()
```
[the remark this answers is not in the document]
AntonOsika: We don't need these, they are set below anyways

AntonOsika (changes requested):
Good job on the PR.


Added my comments. You can immediately commit my proposals in git by clicking commit to batch.

gpt_engineer/main.py
```diff
@@ -21,11 +23,34 @@ def main(
...
+        False,
+        "--improve",
+        "-i",
+        help="Improve code from existing project.",
+    ),
     verbose: bool = typer.Option(False, "--verbose", "-v"),
 ):
     logging.basicConfig(level=logging.DEBUG if verbose else logging.INFO)
 
-    model = fallback_model(model)
```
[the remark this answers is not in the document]
pbharrin: With LangChain integration [this was moved to ai.py](https://github.com/AntonOsika/gpt-engineer/blob/main/gpt_engineer/ai.py#L42) in #512.

gpt_engineer/chat_to_files.py
```diff
@@ -40,3 +41,57 @@ def to_files(chat, workspace):
...
+        with open(file_path, "r") as file:
+            file_data = file.read()
+        if file_data:
+            file_name = os.path.basename(file_path).split("/")[-1]
```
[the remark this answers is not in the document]
pbharrin: We could, however we would have to communicate that whole path with the AI in our prompts.  
My human brain says that's too much information but I'm just making assumptions.  It would be worthwhile to do an experiment with sending the whole path to the AI.  I can do that.

gpt_engineer/main.py
pbharrin: I think it makes sense to just remove the redundant code from your earlier comments.

gpt_engineer/preprompts/implement_on_existing
pbharrin: Aider does not have an explicit "modify existing code" command they have a dialogue to make requested changes.  [This file](https://github.com/paul-gauthier/aider/blob/main/aider/coders/single_wholefile_func_prompts.py#L7) is the start of the dialog.  (Very cool how they get output in the form of Git commits). 

Similarly Mentat does not have an explicit "modify existing code" command but places all commands in a general prompt allowing creation of new files or modifying existing files.  [Here is the prompt.  ](https://github.com/biobootloader/mentat/blob/main/mentat/prompts.py#L14C42-L21C58)

gpt_engineer/chat_to_files.py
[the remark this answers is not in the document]
pbharrin: It works with using the whole path.  The prompt for the return file type needed to be modified as well.

gpt_engineer/main.py
AntonOsika: Not sure I follow here.

You say it's working right now. That sounds like the code is not matching intended behavior because I can't see how these assignments have an effect.

Either code should be deleted or moved, because intended behavior is to use .gpt_eng folder?

PR #580, opened 2023-08-11
Added Gpt-engineer colab notebook

imabutahersiddik:
Run Gpt-engineer on Google Colab in your browser.
It will make user's to try it in their browser before they run it in local system.


AntonOsika:
Good idea but it's not the place to put it etc.

PR #588, opened 2023-08-13
Refactor files for docker

AntonOsika:


AntonOsika:
Would like to get input if this looks ok @karlderkaefer and @ayoubfletcher 🙏 

The goal is to keep the readme.md minimial

Issue #590, opened 2023-08-14
Please update README.md


C0FF33P4G3:
I recently tried using it by following the steps in the README.md file and it does not work, please update the file.

I keep getting this error when i try to export/set the API key

openai.error.AuthenticationError: No API key provided. You can set your API key in code using 'openai.api_key = ', or you can set the environment variable OPENAI_API_KEY=). If your API key is stored
in a file, you can point the openai module at it with 'openai.api_key_path = '. You can generate API keys in the OpenAI web interface. See https://platform.openai.com/account/api-keys for details.

pbharrin:
Thanks, this may have a similar root cause as #589 I will look into that.

redocdm:
experiencing same issue.

pbharrin:
I reproduced this, even with my API key set in .env (following the README) I get this error.  

I will look into getting a fix out.

pbharrin:
If you are still getting the error after my fix, perhaps there is a problem with how you are setting your API key.  
1. are you using the .env file or the env variable?  
2. Can you show exactly how you are setting the API key in 1 without posting your exact key, perhaps replace your key with random keystrokes.

ragazzo-br:
I'm receiving an similar error, but in my case i can put the api keys:
"INFO:openai:error_code=invalid_api_key error_message='Incorrect API key provided: [sk-PSSF*****************************************UYc]. You can find your API key at https://platform.openai.com/account/api-keys.' error_param=None error_type=invalid_request_error message='OpenAI API error received' stream_error=False"

redocdm:
> If you are still getting the error after my fix, perhaps there is a problem with how you are setting your API key.
> 
> 1. are you using the .env file or the env variable?
> 2. Can you show exactly how you are setting the API key in 1 without posting your exact key, perhaps replace your key with random keystrokes.

I am using powershell. Still get same error.
$env:OPENAI_API_KEY = "[my key]"

error: openai.error.AuthenticationError: Incorrect API key provided: [*****************************************]. You can find your API key at https://platform.openai.com/account/api-keys.

I also updated main.py and set the api key directly in there; same error but slightly different message:

pydantic.error_wrappers.ValidationError: 1 validation error for ChatOpenAI
__root__
  Did not find openai_api_key, please add an environment variable `OPENAI_API_KEY` which contains it, or pass  `openai_api_key` as a named parameter. (type=value_error)

AntonOsika:
[...]

Issue #589, opened 2023-08-14
"No API key provided" - altough it is provided in the .env file


LeRobert:
## Expected Behavior

If the OpenAI API key is provided in the .env file, it should be recognized and used.

## Current Behavior

Runtime error message: openai.error.AuthenticationError: No API key provided.

### Steps to Reproduce

1. Set the key in the .env file
2. Run the app with gpt-engineer projects/my-new-project

### Solution

When I added the line `openai.api_key = os.getenv("OPENAI_API_KEY")` to the end of the function `load_env_if_needed()` in the file `main.py`, as well as `import openai` at the beginning of this file _(thanks, engerlina, for reminder)_, the issue was resolved.

engerlina:
You also need to import openai in the main.py file

LeRobert:
> What is the proper place for the .env file?

It works fine if it's in the gpt-engineer's root (same place as .env.template).

redocdm:
does this method only work cloning git versus installing it with pip?

pbharrin:
This method (creating a .env file) should work for either installing with pip or cloning with git.  It would be easier to set the env variable if you installed with pip as you don't have to figure out where to place the .env file.  

I'm going to close this issue as the fix has been merged.

AntonOsika:
I'm a bit confused as of why the PR was necessary.

Is it working better since the fix @LeRobert? Shouldn't having the key as env variable be enough?

If you can explain why importing openai is necessary @engerlina @pbharrin would be great

pbharrin:
I had this same question.  

[The line that is called to set the OpenAI API key](https://github.com/AntonOsika/gpt-engineer/blob/main/gpt_engineer/main.py#L23C5-L23C11) is: 
`openai.api_key = os.getenv("OPENAI_API_KEY")`.  

In order to globally set the OpenAI instance variable `api_key` we first need to have access to the OpenAI singleton.   That is why we need the import statement.

engerlina:
All I know is that I ran into the same issue: Runtime error message: openai.error.AuthenticationError: No API key provided.

And when I updated the code as @LeRobert proposed it worked (except that of course you need to import it).

So once you do both, the code works as expected.

So apologies, if I didn't solve the root cause and I don't know why even with the .env file updated and in the right folders it didn't work before.

LeRobert:
> I'm a bit confused as of why the PR was necessary.
> 
[...]

PR #592, opened 2023-08-15
Fix missing OPENAI_API_KEY

pbharrin:
The PR that enabled .env was breaking.  See issues:  #589 
#590 

I tested this with OPENAI_API_KEY set in both .env and as a bash env variable and it worked.

### DIFF
--- a/gpt_engineer/main.py
+++ b/gpt_engineer/main.py
@@ -6,6 +6,7 @@ from pathlib import Path
 import typer
 
 from dotenv import load_dotenv
+import openai
 
 from gpt_engineer.ai import AI, fallback_model
 from gpt_engineer.collect import collect_learnings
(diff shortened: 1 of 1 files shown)


redocdm:
I looked at your change in main.py. mine from stable version looks different. I added your change and still same error. See below: `import logging

from pathlib import Path

import typer

from dotenv import load_dotenv
import openai

from gpt_engineer.ai import AI, fallback_model
from gpt_engineer.collect import collect_learnings
from gpt_engineer.db import DB, DBs, archive
from gpt_engineer.learning import collect_consent
from gpt_engineer.steps import STEPS, Config as StepsConfig

def load_env_if_needed():
        load_dotenv()
    openai.api_key = os.getenv("OPENAI_API_KEY")

app = typer.Typer()`

AntonOsika:
🙏  @pbharrin

PR #600, opened 2023-08-16
removed file dictionary, and use full path for files in existing code…

pbharrin:
This PR is a cleanup of the Existing Code PR.  In the original implementation there was a dictionary with file names as keys and file full paths as values.  For example: 
main.py --> /Users/boss/gpt-engineer/projects/example/archive/20230816_103041/workspace/main.py

The problem with this is that there could be collision in the dict if files in different folders have the same name.  (For example util.py in multiple folders could cause a collision).  This dict has been completely removed and now the full paths are passed to the AI and returned.

### DIFF
--- a/gpt_engineer/chat_to_files.py
+++ b/gpt_engineer/chat_to_files.py
@@ -51,21 +51,12 @@ def overwrite_files(chat, dbs, replace_files):
 
     files = parse_chat(chat)
     for file_name, file_content in files:
-        # Verify if the file created by the AI agent was in the input list
-        if file_name in replace_files:
-            # If the AI created a file from our input list, we replace it.
-            with open(replace_files[file_name], "w") as text_file:
-                text_file.write(file_content)
-        else:
-            # If the AI create a new file I don't know where to put it yet
-            # maybe we can think in a smarter solution for this in the future
-            # like asking the AI where to put it.
-            #
-            # by now, just add this to the workspace inside .gpteng folder
-            print(
-                f"Could not find file path for '{file_name}', creating file in workspace"
-            )
+        if file_name == "README.md":
+            # creating a README for notes from the AI.  
             dbs.workspace[file_name] = file_content
+        else:
+            with open(file_name, "w") as text_file:
+                text_file.write(file_content)
 
 
 def get_code_strings(input) -> dict[str, str]:
--- a/gpt_engineer/chat_to_files.py
+++ b/gpt_engineer/chat_to_files.py
@@ -43,7 +43,7 @@ def to_files(chat, workspace):
         workspace[file_name] = file_content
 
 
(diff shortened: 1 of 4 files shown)


gpt_engineer/chat_to_files.py
```diff
@@ -51,21 +51,12 @@ def overwrite_files(chat, dbs, replace_files):
...
-            # by now, just add this to the workspace inside .gpteng folder
-            print(
-                f"Could not find file path for '{file_name}', creating file in workspace"
-            )
+        if file_name == "README.md":
+            # creating a README for notes from the AI.  
             dbs.workspace[file_name] = file_content
+        else:
+            with open(file_name, "w") as text_file:
+                text_file.write(file_content)
```
AntonOsika, line 59: Doesn't this do exactly the same thing as `dbs.workspace[file_name] = file_content`?

gpt_engineer/file_selector.py
```diff
@@ -187,7 +188,8 @@ def ask_for_selection(self) -> List[str]:
         elif re.match(regex, user_input):
             try:
                 user_input = (
-                    user_input.replace("", ",") if " " in user_input else user_input
+                    user_input.replace(
+                        "", ",") if " " in user_input else user_input
```
AntonOsika, line 192: Not sure what this is for but shouldn't it be this:

```suggestion
                        " ", ",") if " " in user_input else user_input
```

gpt_engineer/file_selector.py
```diff
@@ -196,10 +198,12 @@ def ask_for_selection(self) -> List[str]:
                         start = int(start_str)
                         end = int(end_str)
                         for num in range(start, end + 1):
-                            selected_paths.append(str(self.selectable_file_paths[num]))
+                            selected_paths.append(
+                                str(self.selectable_file_paths[num]))
```
AntonOsika, line 202: I think this makes the CI break.

Have you configured your pre-commit hooks like described [here](https://github.com/AntonOsika/gpt-engineer/blob/main/.github/CONTRIBUTING.md)?

I think just

`pre-commit install`

AntonOsika (changes requested):
Looks good, I had tw nitpicks though wanna have a look before merging?

AntonOsika:
You can see CI being red above below:

gpt_engineer/chat_to_files.py
pbharrin: yes, it does.  fixed

gpt_engineer/file_selector.py
pbharrin: Fixed.

pbharrin:
The security issue has been addressed, now files paths are sent over relative to the working directory.  I believe I have addressed everything.  

I fixed the pre-checks, thanks for the info I did not know how to do that.

PR #637, opened 2023-08-26
Evals for improve existing code

pbharrin:
This work is motivated by a study of the way our code was being evaluated and used in other tools Auto-GPT, Rift, etc.  
These projects have a modified fork of our code that can be be evaluated in a "headless" fashion with a report generated.  

Since they had a fork of our project, they were not receiving updates to gpt-engineer and the evals were on old branches, which were not being rebased.  This work introduces a method of evaluation which does not require human input, or rather the human input is done ahead of time (in a prompt file).  This is done for the improve code mode, subsequent work will cover the default or new code generation.  

Our output is code, and the code can be tested in one of two ways: 
 - 1. Artifacts in the source code 
 - 2. Performance of the code while being executed.  
 
The first option is more brittle as the AI may have multiple ways of solving a problem.  The second option is less brittle but currently is limited to Python.  With the inclusion of different libraries we can test other languages as we.  Both of these methods are covered in this PR.


pbharrin:
Closing as I have a simpler version of this PR #638

PR #639, opened 2023-08-28
feat: added support for open source models, added instructions to docs

ErikBjare:
Hi @AntonOsika,

Thought I'd give the new WizardCoder models a go, they are pretty impressive!

I was going to try and run the benchmarks, but it was a bit finicky, so I did this instead :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,16 @@ Editing the `preprompts`, and evolving how you write the project prompt, is how 
 
 Each step in `steps.py` will have its communication history with GPT4 stored in the logs folder, and can be rerun with `scripts/rerun_edited_message_logs.py`.
 
+### Running with open source models
+
+You can use gpt-engineer with open source models by using an OpenAI compatible API, such as the one offered by the [text-generator-ui extension `openai`](https://github.com/oobabooga/text-generation-webui/blob/main/extensions/openai/README.md). This can easily be setup with [TheBloke's Runpod template](https://www.runpod.io/console/gpu-secure-cloud?template=f1pf20op0z).
+
+To do so, first set up the API according to the instructions linked above.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -74,6 +74,16 @@ Editing the `preprompts`, and evolving how you write the project prompt, is how
 
 Each step in `steps.py` will have its communication history with GPT4 stored in the logs folder, and can be rerun with `scripts/rerun_edited_message_logs.py`.
 
+### Running with open source models
```
ErikBjare, line 77: This section might go better in the docs (which I just found), although concerned the instructions will simply be buried there. Maybe we could just do it, throw in a link from the README, and call it a day.

README.md
```diff
@@ -74,6 +74,16 @@ Editing the `preprompts`, and evolving how you write the project prompt, is how
...
+
+You can use gpt-engineer with open source models by using an OpenAI compatible API, such as the one offered by the [text-generator-ui extension `openai`](https://github.com/oobabooga/text-generation-webui/blob/main/extensions/openai/README.md). This can easily be setup with [TheBloke's Runpod template](https://www.runpod.io/console/gpu-secure-cloud?template=f1pf20op0z).
+
+To do so, first set up the API according to the instructions linked above.
```
ErikBjare, line 81:
```suggestion
To do so, first set up the API according to the instructions linked above. Then you need to go into the text-generation-webui, go to settings, check the `openai` extension, save. You then need to expose TCP port 5001 in your Runpod config, which will give it an exposed TCP port something like 40125. Then restart your Runpod, and check that the API is live by browsing: http://<public ip>:<port>/v1/models
```

AntonOsika:
Thanks @ErikBjare! 🏃 

Regarding README, it's already too long (want to keep it super short) so I think your idea of linking to docs is the way to go!

Would you be able to update PR?

ErikBjare:
Done! Moved it into the docs, and also added some link to the docs from the README (there wasn't a single one! :scream:)

ErikBjare:
Oops, pushed commits to wrong remote so this PR didn't have all the changes. Addressing in #644

Issue #530, opened 2023-07-13
Using gpt-engineer with Azure OpenAI


RenukaMane:
Hi, I am trying to test gpt-engineer by using Azure OpenAI but I am getting authentication error. I have added all the additional details that are required for the Azure OpenAI like api_base url,  model, etc. in the python file ai.py in the gpt_engineer folder. Am I missing out something can you please help me out with this issue. 

Have set the openAI API key as the windows environmnet variable. Rest all the steps have followed according to the readme file. 

<img width="946" alt="image" src="https://github.com/AntonOsika/gpt-engineer/assets/53396422/d3657e3b-1e49-4f6c-adac-18125ee1f29f">

yanisywd:
Hey @RenukaMane,

Sorry to hear you're having a tough time. Can you double-check your Azure OpenAI API key and make sure your app is picking it up correctly from the environment variable? Also, just to be sure, please confirm the endpoint URL is right. Lastly, keep in mind gpt-engineer might not work seamlessly with Azure OpenAI out of the box. It could be helpful to share more info about the error you're seeing, but remember not to post sensitive stuff like your actual API key

izikeros:
Perhaps you need to set `openai.api_type = "azure"` either via env var or in the code?

SebLz:
I got it working by editing a few lines in ai.py:

![image](https://github.com/AntonOsika/gpt-engineer/assets/190146/5992a061-47bd-4573-84b0-34576e37f9e9)
```
import os

....


# set up azure openai
        if not os.getenv("AZURE_OPENAI_ENDPOINT"):
            raise ValueError("To use Azure OpenAI models please set a "
                            "AZURE_OPENAI_ENDPOINT enviroment variable.") 
        if not os.getenv("AZURE_OPENAI_KEY"):
            raise ValueError("To use Azure OpenAI models please set a "
                            "AZURE_OPENAI_KEY enviroment variable.") 

        openai.api_type = "azure"
        openai.api_base = os.getenv("AZURE_OPENAI_ENDPOINT") 
        openai.api_version = os.getenv("AZURE_OPENAI_API_VERSION")
        openai.api_key = os.getenv("AZURE_OPENAI_KEY")
```

+

![image](https://github.com/AntonOsika/gpt-engineer/assets/190146/27f7eac3-6fd9-4aae-a782-c0e3e6f3a7bf)

+

[...]

PR #640, opened 2023-08-28
Add support for Azure OpenAI Service

niklasfink:
Hi, I'm not sure about the long term strategy how further LLMs might be integrated into gpt-engineer, but for now, I've created an easy Azure OpenAI Service integration.

A couple of people already asked in the issues and since LangChain already supports it, it's quite easy to integrate. #325 #530 

For the future, there might be a better way to integrate various LLMs, but for now I think a simple `--azure` parameter is as easy as it gets.

In addition, I moved the check_consent higher in human_review_input so all the other questions can be skipped.

The Azure OpenAI integration has been tested and works. I couldn't test the OpenAI (default) variant though.

### DIFF
--- a/gpt_engineer/steps.py
+++ b/gpt_engineer/steps.py
@@ -374,7 +374,8 @@ def fix_code(ai: AI, dbs: DBs):
 def human_review(ai: AI, dbs: DBs):
     """Collects and stores human review of the code"""
     review = human_review_input()
-    dbs.memory["review"] = review.to_json()  # type: ignore
+    if review:
+        dbs.memory["review"] = review.to_json()  # type: ignore
     return []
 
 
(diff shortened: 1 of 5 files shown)


pbharrin (commented):
Excellent work!   That was a good idea to move human_review earlier.  One small nit is that [it's recommended](https://peps.python.org/pep-0008/#programming-recommendations) to check for `None` with `is`, so line 377 in steps.py would be: `if review is not None`.

gpt_engineer/steps.py
```diff
@@ -218,11 +218,17 @@ def gen_code_after_unit_tests(ai: AI, dbs: DBs) -> List[dict]:
 def execute_entrypoint(ai: AI, dbs: DBs) -> List[dict]:
     command = dbs.workspace["run.sh"]
 
-    print("Do you want to execute this code?")
+    print()
+    print(
+        colored(
+            "Do you want to execute this code? (y/n)",
```
niklasfink, line 224: I thought this is an important decision and to better see it above the printed execution command below, we paint it in red. Plus, we add the (y/n) which is also used for the other upcoming questions.

PR #659, opened 2023-09-02
Simplify which folders are used

AntonOsika:
This will let folder structure look like so:

```
- .gpt_eng
  - archive
  - memory
- prompt
- [all the generateed code files ("workspace") stored on the top level]
```

It will also stop archiving the workspace – think we should instead version control all the project code via git and let gpt-engineer make commits directly after it generates something.


AntonOsika:
Reopened this, see PR why: https://github.com/AntonOsika/gpt-engineer/pull/669

@pbharrin @hrobbins

Issue #650, opened 2023-09-02
Fix the "improve" prompt to make sure that it generates diffs, and parse and apply those diffs to the existing codebase


AntonOsika:
One way to do this is to write the prompt for gpt-engineer with `-i` flag to annotate each codeblock with one of:

1. `NEW CODE`
2. `REPLACING ONE FUNCTION`

If 1., the generated code can just be written to a new file (or appended to an existing file).

If it is replacing an existing function, we could make sure to find the name of the function that is being replaced using an AST parser (see how [here](https://chat.openai.com/share/71012377-7ebb-47f2-a8fc-7d1bfd4fabe2))

## Why this is necessary
As an example, I tried to use it on the project itself and got a codeblock that was just changing one of the function (so it should not be used to overwrite the entire file)

## How to do it

We can take inspiration from Aider, that generates diffs, or sweep in how they prompt for "<copy_lines>" and [parse the GPT4 output here](https://github.com/sweepai/sweep/blob/e384c9fc3e0278257324c4ce57a888fa64f071b7/sweepai/utils/diff.py#L113)

Should be quite straightforward!

UmerHA:
Hey, I'd take this up. eta a few days

UmerHA:
@AntonOsika I think replacing functions is not quite the right approach, because not every code is neatly made up of functions (eg in python we have code outside functions, classes, nested functions). What do you think about using git style diff files instead?

So we'd ask the LLM to produce something like this:
```
filename.py
second line before code change # used for locating changes
first line before code change
new code
new code
new code
first line after code change
second line after code change
```
I'd ask the LLM to add 2 lines before and after the code change to help us locate the change. The alternative would be asking for line numbers, but as LLMs are notoriously bad with numbers, I'd not go that way.

We would then parse the above output into a git diff file
```
--- a/filename.py
+++ b/filename.py
@@ -1,2 +1,3 @@
-original content
+changed content
unchanged content
```

What do you think?

AntonOsika:
Awesome.

I looked at it took, I think we should first try using [this prompt](https://github.com/paul-gauthier/aider/blob/main/aider/coders/editblock_prompts.py).

It should be possible to use it by doing something like loading EditBlockCoder [here](https://github.com/paul-gauthier/aider/blob/main/aider/coders/editblock_coder.py) and then:

```
coder = EditBlockCoder()  # might need some args to instansiatie
[...]

Issue #665, opened 2023-09-03
Add `project_metadata` the DBs class. Instantiate it in main.py and let its path be `.gpteng`


AntonOsika:
Add a field called `project_metadata` in the DBs class, that is of type DB.

Set project_metadata_path `.gpteng`

Use it when instantiating DBs

Move: memory, archive, file_list.txt, to project_metadata

(Also, separate PR: start writing all_output.txt to memory)

<details open>
<summary>Checklist</summary>

- [X] `gpt_engineer/project_metadata.py` ✅ Commit [`eec2cae`](https://github.com/AntonOsika/gpt-engineer/commit/eec2cae9b893a2bd7b72c99d3cde96e358d70f8b)
<blockquote>• Define the `DB` class in `gpt_engineer/project_metadata.py`.</blockquote>

- [X] `gpt_engineer/db.py` ❌ Failed
<blockquote>• Import the `DB` class from `gpt_engineer.project_metadata`.
• Modify the `DBs` class to include a `project_metadata` field of type `DB`.</blockquote>

- [X] `gpt_engineer/main.py` ❌ Failed
<blockquote>• Import the `DB` class from `gpt_engineer.project_metadata`.
• Instantiate the `project_metadata` field of `DBs` with the path `.gpteng`.</blockquote>

- [X] `gpt_engineer/main.py` ❌ Failed
<blockquote>• Move the `memory`, `archive`, and `file_list.txt` to the `project_metadata` directory.</blockquote>

</details>

ATheorell:
Closing, since solved in #720, right @RareMojo ?

PR #669, opened 2023-09-03
Simplify which folders are used

AntonOsika:
This will let folder structure look like so:

```
- .gpt_eng
  - archive
  - memory
- prompt
- [all the generateed code files ("workspace") stored on the top level]
```

It will also stop archiving the workspace – think we should instead version control all the project code via git and let gpt-engineer make commits directly after it generates something.

Waiting to merge until we have a good way to "merge in" the generated code into the existing code: #650


AntonOsika:
Holding off to merge this until #650 .

After this, we should do #665

ATheorell:
I believe this is obsolete now that #749 has been merged

PR #675, opened 2023-09-05
tkinter install documentation entry

ATheorell:
This documentation addition addresses issue #674. I suggest adding this as a troubleshoot entry in the documentation and not in the readme, since we want to keep the readme short. The problem is of a general nature.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -43,3 +43,20 @@ Once you have a copy of the source, you can install it with:
 
 
 .. _Github repo: https://github.com/AntonOsika/gpt-engineer.git
+
+Troubleshooting
+-------------
+
+For mac and linux system, there are sometimes slim python installations that do not include the gpt-engineer requirement tkinter, which is a standard library and thus not pip installable.
+
+To install tkinter on mac, you can for example use brew:
+
+.. code-block:: console
+
+    $ brew install python-tk
+
+On debian-based linux systems you can use:
+
+.. code-block:: console
+
+    $ sudo apt-get install python3-tk


ATheorell:
Before I start digging, is this issue obvious? I believe that I'm following a very basic workflow and the error occurs before my diff is even considered. GPT4 says this:

The error message you're seeing is related to a step in the GitHub Actions workflow where it attempts to download and extract an action (in this case, actions/checkout@v3). The specific issue is with the tar -xzf command, which is used to extract .tar.gz files.

The return code 2 from the tar command usually indicates an error in the basic usage of the command or a problem with the file that it's trying to extract.

There are a few things to consider or troubleshoot:

    Action Version: As of my last update in September 2021, the latest version of actions/checkout was v2. Ensure that the version v3 exists. It might be a typo or an invalid reference. Check the repository for the latest releases. If v3 doesn't exist, you should use v2.

    Change this:

    yaml

uses: actions/checkout@v3

To this:

yaml

    uses: actions/checkout@v2

    Corrupted File: The archive file being extracted might be corrupted during the download. While this isn't common, network issues or other interruptions can cause this. A re-run of the workflow might resolve this issue if it's a one-time glitch.

    File Permissions: Ensure the runner has the necessary permissions to read/write in the /home/runner/work/_actions/ directory.

    Runner Version: While not as likely, there might be an incompatibility with the runner version (2.308.0 in this case) being used. Ensure that this version of the runner supports all your workflow steps and actions. You might consider trying your workflow on a different version if possible.

If the issue persists after checking these possibilities, you may need to dig deeper into the specific environment and configurations of the runner or consult GitHub's support or community forums for more detailed assistance.

PR #676, opened 2023-09-06
Simplify prompting

AntonOsika:


AntonOsika:
Tested this code so will merge 🏃

PR #691, opened 2023-09-11
Store full file path in workspace

rminchev1:
Previously, only the file name was stored in the workspace. This commit changes the behavior to store the full file path instead. This will help in locating the file more accurately in the workspace. The full path is generated using the os.path.join method.

### DIFF
--- a/gpt_engineer/chat_to_files.py
+++ b/gpt_engineer/chat_to_files.py
@@ -92,8 +92,8 @@ def overwrite_files(chat, dbs):
                 "LAST_MODIFICATION_README.md"
             ] = file_content  # TODO store this in memory db instead
         else:
-            dbs.workspace[file_name] = file_content
-
+            full_path = os.path.join(dbs.input.path, file_name)
+            dbs.workspace[full_path] = file_content            
 
 def get_code_strings(input: DB) -> dict[str, str]:
     """


pbharrin:
Excellent job with this.  
This was the original code, but we changed the code at the last minute to use relative paths instead.  Allow me to explain why.  We thought it may be a security risk.  If we us absolute paths, and the LLM returned the wrong path it could potentially do bad/annoying thins: delete /etc/passwd, or overwrite some other file.  

So using relative paths and not allowing parent directories ".." was our way of keeping this in a sandbox.  What are your thoughts on this?

rminchev1:
It's a good idea to limit what the tool can access. However, I need help finding a way to work with the -i flag (improve feature) simply b/c it always writes LLM responses relative to the workspace folder. Perhaps I need help understanding the usage; with existing code, I expect to pass the files I'm interested in and get them updated/overwritten, not copied to the workspace folder. To me, the improve_code feature is vital because, as we know, the software dev process is iterative & incremental, so I keep using it continuously.

ATheorell:
It sounds like that what you are really looking for is the solution of issue #650  @rminchev1 ? 
Currently, the improve code flag does not have the capability of combining the suggested improvements with the existing code base. Any work on #650 is greatly appreciated.

rminchev1:
@ATheorell , yep correct! IMO this optimization limited the ability to modify files outside the workspace:  https://github.com/AntonOsika/gpt-engineer/commit/26db521788528da23d6a776d6103ae6500117679#diff-5ca0405ef8ecc39a372738948ffd1eec910702b3262b55ca38a0085438ea3b49L95

pbharrin:
The improve code option can modify inside the workspace.  The idea was to put your code in the `projects/` folder.  `projects/` is Git ignored, so you could have another Git repo in there.  @rminchev1 please try putting your existing code there and see if it works.  #650 is for using Git diffs to modify the existing codebase, it would still run into the same problems of modifying code outside of the workspace, but you could undo the changes.  

We really don't have this documented anywhere, and we should.  I think it should be added to the README on the main page.

rminchev1:
@pbharrin, thanks for your feedback. Alright, I still think there's a value to running the tool via the CLI, pointing it to an existing git repo, and letting it work (improving code). With the v0.0.9 release, one can figure out how to plug his repo into the workspace to get it going. If that's the design, it is a good idea to be documented. Otherwise it makes it challenging for newcomers to figure it out at first.

pbharrin:
@rminchev1 that makes total sense.  I will try to get an out edit to the main README today.

pbharrin:
@rminchev1  I have a PR with an updated README at #694 .

rminchev1:
@pbharrin LGTM! Thx.

Issue #683, opened 2023-09-07
Integrate with agentprotocol.ai


AntonOsika:
gpt-engineer is proud to be supporting AEIF, the foundation behind [agent protocol](agentprotocol.ai).

Would be nice to be integrated! 🚀

It's quite straightforward, and gives an **API** and automatic **evaluation** for gpt-engineer:

1. Create a file called api.py
2. Set up a "create task" handler and "access step" handler in there
3. Only support one "step" in the beginning: Generate a codebase
4. See below pseudocode for how to do it

```
from agent_protocol import Agent, Step, Task

def generate_code(task: Task, step: Step) -> Step:
    # only this is gpt-engineer specific
    path = Agent.get_workspace(task.task_id)
    
    dbs = DBs(DB(path / "input"), DB(path / "memory"), ...)
    ai = AI(...)
    for gpte_step in STEPS[StepConfig.simple]:
         gpte_step(ai, dbs)
         
    path = Path("./" + file_path)
    files = chat_to_files(dbs.workspace['all_output.txt'])
    for f, _ in files:
        await Agent.db.create_artifact(
            task_id=task.task_id,
            step_id=step.step_id,
            relative_path= (Path('workspace') / f).parent,
            file_name=Path(f).name
        )
    return step


async def task_handler(task: Task) -> None:
    await Agent.db.create_step(task.task_id, "generate_code")


async def step_handler(step: Step):
    task = await Agent.db.get_task(step.task_id)
    generate_code(task, step)


Agent.setup_agent(task_handler, step_handler).start()
```

AntonOsika:
🚀

PR #693, opened 2023-09-13
Agent Protocol Support

pbharrin:
This PR adds support for the [Agent Protocol](https://agentprotocol.ai/).  #683 There is a file called api.py which implements the protocol and scripts/test_api.py which shows how it can be used.  

Most options are supported except improve code mode.  It is not impossible to support this mode, it will just take some additional code and documentation.

### DIFF
--- a/gpt_engineer/api.py
+++ b/gpt_engineer/api.py
@@ -1,0 +1,60 @@
+import os
+
+from pathlib import Path
+
+from agent_protocol import Agent, Step, Task
+
+from gpt_engineer.db import DB
+from gpt_engineer.main import main
+
+
+async def task_handler(task: Task) -> None:
+    """task_handler should create initial steps based on the input.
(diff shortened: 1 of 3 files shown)


ATheorell:
Agent Protocol Support
The test script runs very smoothly.
I spent quite a long time, trying to get gpt-engineer to communicate with https://github.com/Significant-Gravitas/Auto-GPT/tree/master/benchmark via the protocol, while running locally. Took a long time to get their code to run at all and tried many commits. Commit  Fix API Mode #5209 (only 5 hours old) of AutoGPT runs and sends requests to gpt-engineer, but currently, only errors come out. 
I merged now, so that people can already start using it, but once the AutoGPT development stabilizes, we should probably debug why all tests fail currently.

AntonOsika:
🚀

PR #697, opened 2023-09-14
Handle undefined additional_information in agent protocol api implementation

ATheorell:
Got errors when using the API for the AutoGPT benchmarks, because the GETs dont include the additional_information property. This PR fixes that issue.


pbharrin (changes requested):
Good catch, I think we can reduce this to a one line edit if we make line 44 like this: 
` additional_properties=task.additional_input.__root__
        if task.additional_input is not None
        else {},`
        
That way `step.additional_properties` will be an empty dictionary and we can keep the rest of the code the same.

PR #698, opened 2023-09-15
Adding official agent protocol test to ci + github badge

ATheorell:


ATheorell:
Is there an obvious reason why neither the ci.yaml, nor validate_agent_protocol.yml are run in the checks? They are the whole purpose with this PR... @pbharrin

ATheorell:
I can reproduce this failure locally and it is because the OPENAI_API_KEY is not defined. Long term, it would be nice to have some kind of dummy model that could be used. 

For now, the easy fix is to add the api key to github as a secret, once that is done, the action script should already be reading from it. Would be great if you could do this @AntonOsika https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions (I don't have permission). Call the secret OPENAI_API_KEY 

And while your at it, feel free to add a codecov.io token as a secret too, which will enable us to monitor test coverage.

ATheorell:
This is currently failing because pip install agent-protocol is down. Waiting til they get it up again.

ATheorell:
Now all tests are passing. I took the freedom to make a further update and use tempdir as the default for hosting the tasks. Let me know if this has any clear disadvantages.
We bypass the lack of a api key in the online test by catching the authentication error and printing a message. Not sure if this is agood long term solution.

pbharrin (commented):
LGTM, very professional cleanup.

PR #704, opened 2023-09-16
file_selector path as input path

lukaspetersson:
To select files for improvement faster, the inital path of the file selector should be the input path.

This is especially important if using the CLI option as the printed tree can become very large otherwise.

### DIFF
--- a/gpt_engineer/file_selector.py
+++ b/gpt_engineer/file_selector.py
@@ -270,10 +270,10 @@ def ask_for_files(db_input) -> None:
 
     if selection_number == 1:
         # Open GUI selection
-        file_path_list = gui_file_selector()
+        file_path_list = gui_file_selector(db_input.path)
     elif selection_number == 2:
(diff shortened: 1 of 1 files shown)


ATheorell:
Cannot judge whether there are any unforseen drawbacks of this. Input @pbharrin @UmerHA ?

UmerHA:
> Cannot judge whether there are any unforseen drawbacks of this. Input @pbharrin @UmerHA ?

Hey, I don't see any drawback. Both changed functions (`gui_file_selector` and `terminal_file_selector`) still return a list of filenames, so it should be fine.

AntonOsika:
I like this.

Makes me think we should refactor the core logic and CLI UX specific things to make it easier to take decisions on what makes sense from UX perspective, and how the core building blocks should work.

AntonOsika:
See #718, tagging @lukaspetersson for visibility

PR #706, opened 2023-09-16
Print total api costs

lukaspetersson:
Cost conscious users would want to monitor their api usage.

Prints the total cost at the end:
```
...
If you have time, please explain what was not working (ok to leave blank)

Total api cost: 0.30204
```

### DIFF
--- a/gpt_engineer/ai.py
+++ b/gpt_engineer/ai.py
@@ -9,6 +9,7 @@ from typing import List, Optional, Union
 import openai
 import tiktoken
 
+from langchain.callbacks.openai_info import MODEL_COST_PER_1K_TOKENS
 from langchain.callbacks.streaming_stdout import StreamingStdOutCallbackHandler
 from langchain.chat_models import AzureChatOpenAI, ChatOpenAI
 from langchain.chat_models.base import BaseChatModel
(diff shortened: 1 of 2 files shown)


gpt_engineer/ai.py
```diff
@@ -271,6 +271,27 @@ def format_token_usage_log(self) -> str:
...
+        """
+        Return the total cost in USD of the api usage.
+
+        Returns
+        -------
+        float
+            Cost in USD.
+        """
+        prompt_price = 0.03
+        completion_price = 0.06
```
haseeb-heaven, line 284: Prices should not be hardcoded what if they change prices in future ?

lukaspetersson: I agree that it is not ideal, but getting it with an api call would be uneccesary overhead.
[Langchain also hardcodes](https://github.com/langchain-ai/langchain/blob/116cc7998cac563894bdf0037259db767882b55c/libs/langchain/langchain/callbacks/openai_info.py#L7C1-L7C25)

But maybe it is better to get if from langchain as they might be more reliable in updating it.

lukaspetersson: Updated to get the price from langchain.

haseeb-heaven: I still think its better to get prices for each model from the API or use some cache mechanism to store the prices and keep on eye on API prices changes.

ATheorell:
I think this is a great idea!

Would you mind adding a dollar sign on main.py, line  99, so that it becomes:

print("Total api cost: $ ", ai.usage_cost())

then I will merge.

PR #714, opened 2023-09-18
More clear readme instructions

lukaspetersson:
As a new user, the instructions confused me.
It gave me a false sense of how gpt-engineer handles paths.

Note: this assumes that code for improvement does not have to be in the `projects/` folder, is this true?


ATheorell:
We should develop this further after merging #650 to also include which output is generated where.

Issue #708, opened 2023-09-17
Prompt too complex - Clarification makes it worse


lukaspetersson:
## Expected Behavior

gpt-engineer should only ask when clarification is really needed. Otherwise it just ads to an already complex prompt.

## Current Behavior

It allways asks a bunch of questions even if the information is provided in the prompt.

## Failure Information

I tried to make a script that reads a redis database and prints the information in a table. I provided the full information to do the task, but it still asked questions (which could be answered by the info in the prompt). The double-info made the prompt too long and confusing and the output used localhost even if the prompt told it not to. On the otherhand, putting the prompt in chatGPT worked first try.

### Failure Logs

Prompt:
```
Create a Python script that fetches data from a redis database. The key of the hash is usersInfo. The format of the data in the hash is {userId}_{attribute} where the attributes are: [name, last_login, email, chats_count]. You can get the data of the usersInfo with hgetall and you can get all the userIds with the keys * command, each user has a key of the format "user:chat:*". The result of querying the database will be a object where the data is in a list under the key "result".

Use prettytable library to print a table with a row for each user and a column for each attribute.

use fetch to get the data from the databse, it is not hosted in localhost:
fetch("https://big-stag-42761.kv.vercel-storage.com/set/user_1_session/session_token_value", {
  headers: {
    Authorization: "Bearer ****"
  }
}).then(response => response.json())
  .then(data => console.log(data));

```

Output:
```
Summary of areas that need clarification:

1. The specific Python libraries or modules to use for interacting with the Redis database.
2. The structure of the data fetched from the Redis database.
3. The process of fetching data from the non-localhost database using the provided URL.
4. The use of the 'fetch' function in the context of a Python script.
5. The process of using the 'prettytable' library to print the fetched data.
6. The format of the final output object and how to structure it.
7. The use of the 'Authorization' header in the fetch request.

Clarifying question:

Could you please clarify how to use the 'fetch' function in the context of a Python script, as it seems to be JavaScript syntax?
(answer in text, or "c" to move on)
c

(letting gpt-engineer make its own assumptions)

Assumptions:

1. Since 'fetch' is a JavaScript function, I assume that we will use the 'requests' library in Python to make the HTTP request to the provided URL.
2. The 'Authorization' header will be included in the headers of the 'requests' function in Python.
3. The Redis database can be interacted with using the 'redis' library in Python.
[...]

PR #715, opened 2023-09-18
New clarification prompt

lukaspetersson:
#708 and #703 

tldr: clarification flow is confusing for user and adds unnecessary information that confuses GPT.

Note: does not make the list of clarifications anymore.

I would love feedback on whether this could cause unexpected consequences.

### DIFF
--- a/gpt_engineer/steps.py
+++ b/gpt_engineer/steps.py
@@ -98,13 +98,11 @@ def clarify(ai: AI, dbs: DBs) -> List[Message]:
             print()
             return messages
 
-        user_input += (
-            """
+        user_input += """
             \n\n
             Is anything else unclear? If yes, ask another question.\n
             Otherwise state: "Nothing to clarify"
             """
-        )
 
     print()
     return messages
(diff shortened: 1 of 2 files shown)


gpt_engineer/steps.py
```diff
@@ -76,12 +76,12 @@ def clarify(ai: AI, dbs: DBs) -> List[Message]:
...
         msg = messages[-1].content.strip()
 
-        if msg == "Nothing more to clarify.":
+        if "nothing to clarify" in msg.lower():
             break
 
         if msg.lower().startswith("no"):
-            print("Nothing more to clarify.")
             break
+            print("Nothing to clarify.")
```
AntonOsika, line 84: Dead line?

lukaspetersson: I apologize, fixed now

AntonOsika:
I think clarify is very much a UX questions.

Are you confident this doesn't make it unstable/degrade behavior/etc?

We could also just go for #717 instead

lukaspetersson:
> I think clarify is very much a UX questions.
> 
> Are you confident this doesn't make it unstable/degrade behavior/etc?
> 
> We could also just go for #717 instead

#717 just hides clarify, but it still leaves all the problems with it if user decides to use it. 

I ran it a couple of times (twice on 2 different cases) and I very much prefered the UX over the current version. It results in a shorter total prompt, which is beneficial I think, but this is just speculation and intuition.

AntonOsika:
Cool let’s go

PR #717, opened 2023-09-18
Stop clarifying per default, to make runs faster, put it behind --steps clarify flag

AntonOsika:


lukaspetersson:
LGTM.

Regarding the philosophy of the steps, I think it would be better if all "step functions" have the same effect. I.e. all should return messages and call to_files(). clarify() does not call to_files for example and is only ever run if before gen_clarified_code(). Should these not merge then for consistency?

lukaspetersson (approved):

PR #720, opened 2023-09-19
Add project_metadata as .gpteng and relocate files

RareMojo:
Added new DB called `project_metadata` with the path of `.gpteng`.
Moved memory, archive, and `file_list.txt` to `.gpteng`.

issue #665


pbharrin:
- Excellent work!  LGTM.  There are two eval scripts in evals/ that will put file_list.txt elsewhere but I can fix those.

PR #723, opened 2023-09-19
Fix -i flag after metadata was introduced

AntonOsika:


AntonOsika:
Merging!

FYI @RareMojo and @pbharrin :)

Issue #682, opened 2023-09-07
Explicit "code review" step + self healing code


AntonOsika:
The proposal is most easily explained with specific steps to carry out:

After the generation finishes, we would:
1. Ask the LLM what a QA person would manually test based on the project specification ("does snake move", "does the snake change direction when key is pressed" etc)
2. For each QA tsak, ask the LLM if the generated code is expected to pass the QA task just by inspecting the code, or if something seems to be broken
3. Advanced: Let the LLM optionally output _code blocks_ that overwrites files if it detects that something is broken in the QA, and it can fix it

ATheorell:
I think point 3 should be left untouched in this issue for now, since such functionality is underway in #650 and #727 .

yskolnick:
I am going to work on this on Monday or Tuesday.

TheoMcCabe:
@yskolnick Hows this going are you still working on it? This would be useful for the hackathon

yskolnick:
Hey! Apologies.  I got distracted with other work.  I will work on it in next few days.

AntonOsika:
Great @yskolnick. Looking forward to see how this works over the weekend!

PR #727, opened 2023-09-20
Selfheal

pbharrin:
A working version of self-healing code.  This is a single step which can be added to the end of other chains using one of the `gen`s.    This is part of a solution to #682 

Attempts to execute the code from the entry point and if it fails, sends the error output back to the AI with instructions to fix.

This code will make `MAX_SELF_HEAL_ATTEMPTS` to try and fix the code before giving up.

This makes the assuption that the previous step was `gen_entrypoint`, this code could work with `simple_gen`, or `gen_clarified_code` as well.  We just need a way to specify which step was used to create the code.  

Additionally similar code can be used to execute unit tests and run them.

### DIFF
--- a/gpt_engineer/cli/main.py
+++ b/gpt_engineer/cli/main.py
@@ -160,6 +160,7 @@ def main(
         StepsConfig.USE_FEEDBACK,
         StepsConfig.EVALUATE,
         StepsConfig.IMPROVE_CODE,
+        StepsConfig.SELF_HEAL,
     ]:
         archive(dbs)
         load_prompt(dbs)
--- a/gpt_engineer/core/steps.py
+++ b/gpt_engineer/core/steps.py
@@ -50,6 +50,8 @@ import re
 import subprocess
 
 from enum import Enum
+from platform import platform
+from sys import version_info
 from typing import List, Union
 
 from langchain.schema import AIMessage, HumanMessage, SystemMessage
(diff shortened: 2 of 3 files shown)


gpt_engineer/steps.py
```diff
@@ -19,9 +21,22 @@
...
+
 Message = Union[AIMessage, HumanMessage, SystemMessage]
 
 
+def get_platform_info():
+    """Returns the Platform: OS, and the Python version.
+    This is used for self healing.
+    """
+    v = version_info
+    a = f"Python Version: {v.major}.{v.minor}.{v.micro}"
```
fabhed, line 35: Thinking this could create confusion if the python version that's running gpt-engineer is not the same version running the project from run.sh

It would also not be relevant if the project is written in another language.

Not sure how to best solve this, if it makes noticeable improvements to the flow. Maybe we could make run.sh include some dynamic version output, that we either save to a file or print to stdout, which this flow could read from.

pbharrin: Thanks for taking a look at my PR.  

This does make a noticeable improvement to the self healing code.  The AI will propose many solutions such a pre-compiled binary, but if the AI doesn't know the OS and Python version these will not work. 

I think the most ideal solution would be to ask the AI to create a version printing executable based on `run.sh`.  We would have to weigh if the added overhead of the roundtrip to the AI and the added time is worth it, and this would assume the AI gets it right 100% of the time.  

To me it is not worth it as every `run.sh` I've seen has used the default version of Python in the venv.  We can ask @AntonOsika

ATheorell:
@whatactuallyis no it is not yet merged to master. If you want to experiment with it, I recommend you merge it locally (or just copy the relevant part since the merge is non-trivial).

gpt_engineer/steps.py
```diff
@@ -380,6 +395,72 @@ def fix_code(ai: AI, dbs: DBs):
...
+        except subprocess.TimeoutExpired:
+            timed_out = True
+            print("The process hit a timeout before exiting.")
+
+        # get the result and output
+        # step 2. if the return code not 0, package and send to the AI
+        if p.returncode != 0 and not timed_out:
+            print("run.sh failed.  Let's fix it.")
+
+            # pack results in an AI prompt
```
TheoMcCabe, line 438: So this PR doesnt actually doself heal right its just the framework in place? is the idea to combine this with the new improve code?

ATheorell: No its does self-heal, but it rewrites the whole program in every iteration. I see it as a first pass.
The advantage with having an incremental self-heal using --improve would be that it eats less time and token.

TheoMcCabe: Nice- whats the plan for merging? I'd suggest if shall we just get the conflicts resolved and merge it without adding to the default steps for now? Then we can continue to improve in main branch

TheoMcCabe: Im happy to do this today if you agree

pbharrin: I will work on getting the conflicts resolved.

PR #728, opened 2023-09-22
Fix: AI-generated file path issues with -i

RareMojo:
Remove `workspace\` from the `file_list.txt` context to prevent nesting another workspace folder.

Could do this a dozen ways. 
Core issue: GPT would state filenames as `workspace\...` causing new file creation and no overwrites.

### DIFF
--- a/gpt_engineer/chat_to_files.py
+++ b/gpt_engineer/chat_to_files.py
@@ -141,4 +141,6 @@ def format_file_to_input(file_name: str, file_content: str) -> str:
     {file_content}
     ```
     """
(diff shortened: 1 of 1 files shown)


ATheorell:
Can you make a solution that does this, but:

- works on all operating systems
- only targets directories that are called exactly workspace (the current one can distort cases when workspace is a substring in an actual directory)
- Only removes the first occurrence of a folder called workspace in the path

I may be missing something :)

ATheorell:
I imagine this is  obsolete, now that the concept of workspace has been removed. Happy to open again in case I'm wrong

PR #733, opened 2023-09-22
Lite mode - run without pre prompts for execution

lukaspetersson:
Trying to adress parts of #708

The -l CLI flag runs gpt-engineer in lite mode.
It only includes the main prompt and the fileformat.
The idea is that it should be like normal chatGPT but without copy-paste work.

I ran a few tests, unsurprisingly it is faster and cheaper.
I also experience that the output is better.

Examples:

```
gpt-engineer txtfile -l
hello_world.txt
\```
Hello world
\```Total api cost: $  0.00438
```

```
gpt-engineer db_analytics -l 
Here is the Python script that fetches data from a Redis database and prints it in a table format using the prettytable library. 

Please note that the Python script uses the requests library to fetch data from the database, as the fetch API is not available in Python. 

Also, the Python script assumes that the Redis database returns JSON data. 

main.py
\```python
import requests
import json
from prettytable import PrettyTable

def fetch_data_from_redis():
    url = "https://big-stag-42761.kv.vercel-storage.com/set/user_1_session/session_token_value"
    headers = {
        "Authorization": "Bearer ****"
    }
    response = requests.get(url, headers=headers)
    data = response.json()
    return data

def get_user_ids(data):
    user_ids = []
    for key in data.keys():
        user_id = key.split("_")[0]
        if user_id not in user_ids:
            user_ids.append(user_id)
    return user_ids

def get_user_info(data, user_ids):
    user_info = {}
    for user_id in user_ids:
        user_info[user_id] = {
            "name": data.get(f"{user_id}_name"),
            "last_login": data.get(f"{user_id}_last_login"),
            "email": data.get(f"{user_id}_email"),
            "chats_count": data.get(f"{user_id}_chats_count"),
        }
    return user_info

def print_table(user_info):
    table = PrettyTable()
    table.field_names = ["User ID", "Name", "Last Login", "Email", "Chats Count"]
    for user_id, info in user_info.items():
        table.add_row([user_id, info["name"], info["last_login"], info["email"], info["chats_count"]])
    print(table)

def main():
    data = fetch_data_from_redis()
    user_ids = get_user_ids(data)
    user_info = get_user_info(data, user_ids)
    print_table(user_info)

if __name__ == "__main__":
    main()
\```
```

### DIFF
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -206,4 +206,3 @@ Functions
     steps.setup_sys_prompt
     steps.setup_sys_prompt_existing_code
     steps.use_feedback
-
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -187,7 +187,6 @@ Functions
     :toctree: steps
 
     steps.assert_files_ready
-    steps.basic_gen
     steps.clarify
     steps.curr_fn
     steps.execute_entrypoint
@@ -205,4 +204,5 @@ Functions
     steps.set_improve_filelist
     steps.setup_sys_prompt
     steps.setup_sys_prompt_existing_code
+    steps.simple_gen
     steps.use_feedback
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -23,6 +23,23 @@ Functions
     ai.get_tokenizer
     ai.serialize_messages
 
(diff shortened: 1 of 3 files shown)


AntonOsika:
Looks good, merging!

Let's encourage people to use it and consider making it the default

PR #737, opened 2023-09-23
Remove experimental TDD step configs

AntonOsika:
Solves #656 

Kept them on the experimental branch


gpt_engineer/steps.py
```diff
@@ -172,23 +134,6 @@ def gen_clarified_code(ai: AI, dbs: DBs) -> List[dict]:
     return messages
 
 
-def gen_code_after_unit_tests(ai: AI, dbs: DBs) -> List[dict]:
```
lukaspetersson, line 175: Removing this causes unit test test_collect.py to fail. Delete the test?

PR #740, opened 2023-09-24
Load env variable handling & Project preprompt enhancement

saschalalala:
Hey everyone 👋, 

thanks for the great project. I found some things that I believe can benefit the project.

# env file loading

I use pyenv and pyenv-virtualenv and I noticed that my pip installed gpt-engineer always tries to load the .env from its site-packages directory and searches up to the root. My project is located somewhere else, so it did not find my .env. I introduced a small change that as a fallback, if the .env file isn't found before, will attempt to find it in the current working directory. (c834e6998e4e96502e7fa3cb42d2beb04b5cf4d3 and ac4cac62dcc8ae1eb989cd648f943f831eb35e2a)

# preprompt management

I noticed that the preprompts are part of the project itself and a change inside my venv directory would a) be overwritten by an update and b) be global. I wanted to have project specific preprompts so I implemented them. I noticed that the `DB` class cannot handle kind of a "merged" path so that only the files that I want to overwrite would come from the project specific directory so I decided to copy all existing files to the project so that when the preprompts would be updated in the future (inside gpt-engineer itself) they would be copied over to the project directory.

In order to test this feature and because I don't know every implication of the preprompts yet, I changed all of them inside my example project to say `You do not code, you only shout back at people` and it looks like this:

```
gpt-engineer --use-project-preprompts projects/example                                               
I'M SORRY, BUT I CAN'T HELP YOU WITH THAT. I DON'T CODE, I ONLY SHOUT BACK AT PEOPLE!It seems like there's a misunderstanding. Could you please provide the actual information about the codebase?
```

```
gpt-engineer projects/example           
To implement the game Snake in Python using the MVC (Model-View-Controller) design pattern, we will need several classes and files. The game will be controlled using the keyboard. Here are the core classes and their purposes: …
```

# Future steps if this PR is accepted

- [ ] doc enhancement
- [ ] tests

### DIFF
--- a/gpt_engineer/main.py
+++ b/gpt_engineer/main.py
@@ -20,6 +20,9 @@ app = typer.Typer()  # creates a CLI app
 def load_env_if_needed():
     if os.getenv("OPENAI_API_KEY") is None:
         load_dotenv()
+    if os.getenv("OPENAI_API_KEY") is None:
+        # if there is no .env file, try to load from the parent directory
+        load_dotenv(dotenv_path=os.getcwd() + "/.env")
     openai.api_key = os.getenv("OPENAI_API_KEY")
 
 
--- a/gpt_engineer/main.py
+++ b/gpt_engineer/main.py
@@ -92,7 +92,7 @@ def main(
     preprompts_path = Path(__file__).parent / "preprompts"
 
     if use_project_preprompts:
-        project_preprompts_path = workspace_path / "preprompts"
+        project_preprompts_path = input_path / "preprompts"
         if not project_preprompts_path.exists():
             project_preprompts_path.mkdir()
 
--- a/gpt_engineer/main.py
+++ b/gpt_engineer/main.py
@@ -21,7 +21,7 @@ def load_env_if_needed():
     if os.getenv("OPENAI_API_KEY") is None:
         load_dotenv()
     if os.getenv("OPENAI_API_KEY") is None:
-        # if there is no .env file, try to load from the parent directory
+        # if there is no .env file, try to load from the current working directory
         load_dotenv(dotenv_path=os.getcwd() + "/.env")
     openai.api_key = os.getenv("OPENAI_API_KEY")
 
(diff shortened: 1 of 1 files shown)


gpt_engineer/main.py
```diff
@@ -20,6 +20,9 @@
 def load_env_if_needed():
     if os.getenv("OPENAI_API_KEY") is None:
         load_dotenv()
+    if os.getenv("OPENAI_API_KEY") is None:
+        # if there is no .env file, try to load from the current working directory
+        load_dotenv(dotenv_path=os.getcwd() + "/.env")
```
ATheorell, line 25: Please use:
load_dotenv(dotenv_path=os.path.join(os.getcwd(), ".env"))
instead of direct concatenation.

ATheorell (changes requested):
I believe this increases useability and reproducibility.

see my comment on line 25

gpt_engineer/main.py
saschalalala: Done in the last commit, thanks for the hint.

AntonOsika (commented):
Thanks for the PR @saschalalala !

To make it a nice legacy of yours, would you consider refactoring it to make the main() file?

Could set line 109 to preprompts=DB(preprompt_path(use_custom_preprompts))

(`use_custom_preprompts: bool` flag, `preprompt_path(use_custom_preprompts: bool)` is a function)

gpt_engineer/main.py
```diff
@@ -80,13 +89,24 @@ def main(
...
+
+    if use_project_preprompts:
+        project_preprompts_path = input_path / "preprompts"
+        if not project_preprompts_path.exists():
+            project_preprompts_path.mkdir()
+
+        for file in preprompts_path.glob("*"):
+            if not (project_preprompts_path / file.name).exists():
+                (project_preprompts_path / file.name).write_text(file.read_text())
+        preprompts_path = project_preprompts_path
```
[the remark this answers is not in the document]
saschalalala: Will do

saschalalala:
Of course :) 

I also already started writing some docs for that feature. Will create a new PR after the weekend.

PR #745, opened 2023-09-27
Update README.md

aditeyapatakoti:
redid the grammar that I... well redid 💀


ATheorell:
This contribution is too minor to be integrated. If you feel like going through the docs and checking if the technical instructions (what really matters much) are up to date and fixing the problems there I will merge.

PR #753, opened 2023-10-02
Update Dockerfile

woutervanranst:
Bumping python version to address https://github.com/AntonOsika/gpt-engineer/issues/746#issuecomment-1743152184


ATheorell:
Thanks!

PR #768, opened 2023-10-05
Update installation.rst

aditeyapatakoti:


ATheorell:
I don't think this is necessary. gpt-engineer requires 3.8<=python<4.0 which is pretty inclusive and not the latest.

ATheorell:
That's honest! I suggest you make a real contribution :)

PR #769, opened 2023-10-05
Fix: Crashes after entering additional questions

### LINKED ISSUE #754: the code fails after giving additional information at the questions.
## Policy and info
 - Maintainers will close issues that have been stale for 14 days if they contain relevant answers.
 - Adding the label "sweep" will automatically turn the issue into a coded pull request. Works best for mechanical tasks. More info/syntax at: https://docs.sweep.dev/

## Expected Behavior
That I get some feedback, e.g. code that I can use.

## Current Behavior
Fails after the additional questions.

## Failure Information
Nothing more to clarify.

  File "/Users/tom/Library/Python/3.9/bin/gpt-engineer", line 8, in <module>
    sys.exit(app())

  File "/Users/tom/Library/Python/3.9/lib/python/site-packages/gpt_engineer/main.py", line 96, in main
    messages = step(ai, dbs)

  File "/Users/tom/Library/Python/3.9/lib/python/site-packages/gpt_engineer/steps.py", line 192, in gen_clarified_code
    messages = AI.deserialize_messages(dbs.logs[clarify.__name__])

  File "/Users/tom/Library/Python/3.9/lib/python/site-packages/gpt_engineer/ai.py", line 216, in deserialize_messages
    return list(messages_from_dict(json.loads(jsondictstr)))  # type: ignore

  File "/Users/tom/Library/Python/3.9/lib/python/site-packages/langchain/schema/messages.py", line 351, in messages_from_dict
    return [_message_from_dict(m) for m in messages]

  File "/Users/tom/Library/Python/3.9/lib/python/site-packages/langchain/schema/messages.py", line 351, in <listcomp>
    return [_message_from_dict(m) for m in messages]

  File "/Users/tom/Library/Python/3.9/lib/python/site-packages/langchain/schema/messages.py", line 331, in _message_from_dict
    return AIMessage(**message["data"])

  File "/Users/tom/Library/Python/3.9/lib/python/site-packages/langchain/load/serializable.py", line 90, in __init__
    super().__init__(**kwargs)

  File "/Users/tom/Library/Python/3.9/lib/python/site-packages/pydantic/v1/main.py", line 341, in __init__
    raise validation_error

pydantic.v1.error_wrappers.ValidationError: 1 validation error for AIMessage
is_chunk
  unexpected value; permitted: False (type=value_error.const; given=True; permitted=(False,))

python --version
Python 3.11.6

chatgpt API, 3.5-turbo

Possibly related waring/issue I get is:
Users/tom/Library/Python/3.9/lib/python/site-packages/urllib3/__init__.py:34: NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'LibreSSL 2.8.3'. See: https://github.com/urllib3/urllib3/issues/3020

### Steps to Reproduce

If possible, provide detailed steps for reproducing the issue.

1. I have a prompt file (no extension) in a folder
2. I run gpt-engineer folder
3. I get additional questions
4. After that, it fails everytime, tried 3 different folders with different prompts, I once skipped the questions, other times I answered them all.. fail everytime.

### Failure Logs

Any relevant log snippets or files here.

pralad-p:
Fixes #754. 

### Cause: 
From a recent Langchain update, it seems the AIMessage object has a much stricter schema which only allows `is_chunk` attributes to have a `False` value. However, in gpt_engineer, the deserialization of `clarify` logs implicitly sets `is_chunk` to `True`. 

### Proposed fixes:
1. Either go to Langchain (or Pydantic which they use internally for model validation of their types) and loosen their requirements for an AIMessage object model. They may or may not agree with my proposal. 
2. Make it explicit that when message objects are deserialized from the  JSON string, that their `is_chunk` attribute can never be `True`. 

### Containment/Fix: 
I go with **Fix 2**. 
This commit deserializes and stores an intermediate list, from which we explicitly set all `is_chunk` values to `False`. Since the Langchain messages_from_dict(..) method anyways expect a Message object with an `is_chunk` property of `False`, we don't have the effect of hotfixing this property, but rather add an additional check to make sure Messages have valid properties (like this is_chunk property for e.g.).

### DIFF
--- a/gpt_engineer/core/ai.py
+++ b/gpt_engineer/core/ai.py
@@ -342,12 +342,11 @@ class AI:
         """
         data = json.loads(jsondictstr)
         # Modify implicit is_chunk property to ALWAYS false
-        # since Langchain's Message schema is more strict
-        modified_data = [
+        # since Langchain's Message schema is stricter
+        prevalidated_data = [
             {**item, "data": {**item["data"], "is_chunk": False}} for item in data
         ]
-        jsondictstr = json.dumps(modified_data)  # assign to original JSON string
-        return list(messages_from_dict(json.loads(jsondictstr)))  # type: ignore
+        return list(messages_from_dict(prevalidated_data))  # type: ignore
 
     def update_token_usage_log(
         self, messages: List[Message], answer: str, step_name: str
--- a/gpt_engineer/core/ai.py
+++ b/gpt_engineer/core/ai.py
@@ -340,6 +340,11 @@ class AI:
         List[Message]
             The deserialized list of messages.
         """
+        data = json.loads(jsondictstr)
+        # Modify implicit is_chunk property to ALWAYS false
+        # since Langchain's Message schema is more strict
+        modified_data = [{**item, 'data': {**item['data'], 'is_chunk': False}} for item in data]
+        jsondictstr = json.dumps(modified_data)
         return list(messages_from_dict(json.loads(jsondictstr)))  # type: ignore
 
     def update_token_usage_log(
--- a/gpt_engineer/core/ai.py
+++ b/gpt_engineer/core/ai.py
@@ -343,8 +343,10 @@ class AI:
         data = json.loads(jsondictstr)
         # Modify implicit is_chunk property to ALWAYS false
         # since Langchain's Message schema is more strict
-        modified_data = [{**item, 'data': {**item['data'], 'is_chunk': False}} for item in data]
-        jsondictstr = json.dumps(modified_data)
+        modified_data = [
+            {**item, "data": {**item["data"], "is_chunk": False}} for item in data
+        ]
+        jsondictstr = json.dumps(modified_data)  # assign to original JSON string
         return list(messages_from_dict(json.loads(jsondictstr)))  # type: ignore
 
     def update_token_usage_log(


ATheorell:
Thanks for picking this up. Just answer my minor question and I will merge.

pralad-p:
> Thanks for picking this up. Just answer my minor question and I will merge.

Sorry, I missed your question?

ATheorell:
Is the call on line 349 really necessary? Can't you just directly put modified_data in the return and omit the json.loads statement on the following line?
It looks like we are serializing and deserializing on two consecutive lines.

ATheorell:
Thanks, great that you dug out the error.

pralad-p:
Done ✅ I would also recommend to put this in a minor release for its PyPi package soon (I'm not sure what your release cycles are like), because this breaks the application whenever the user is asked for clarification. Even if the user skips.

pralad-p:
> Thanks, great that you dug out the error.

No problem, glad I could help 😄

bhanuunrivalled:
gpt-engineer            0.0.9

i have this problem tooo

in which version this is solved?

ATheorell:
So, it is solved on commit https://github.com/AntonOsika/gpt-engineer/commit/d37feec9bb29f1c81ceead9bea584d9c9e1f41d2
We currently do not update the version in each commit, so many commits are called 0.0.9. If you refer to the version on pypi, this bug still persists.

PR #772, opened 2023-10-07
Fix preprompt path

AntonOsika:


AntonOsika:
Fixes bug with not finding preprompts

We need tests!

chrisivogt:
> Fixes bug with not finding preprompts
> 
> We need tests!

I was able to use preprompts after this commit!

PR #777, opened 2023-10-09
Bugfix/improve mode not changing files

TheoMcCabe:
I noticed that if i created a project with gpt engineer - then attempted to improve it this didn't work. There was a bug which meant changes to files were being written to workspace/workspace. To fix i have set the workspace to the root folder in improve mode. However this meant i had to move the all_output.txt files to the memory folder otherwise the improve steps couldn't find it.  As this was listed in the code as TODO that didn't seem to be too much of a liberty for fixing this bug.

### DIFF
--- a/tests/test_chat_to_files.py
+++ b/tests/test_chat_to_files.py
@@ -3,6 +3,18 @@ import textwrap
 from gpt_engineer.core.chat_to_files import to_files
 from gpt_engineer.core.db import DB, DBs
 
+from typing import Union
+from pathlib import Path
+
+class StubDB(DB):
+    def __init__(self, path: Union[str, Path]):
+        self.data = {}
+
+    def __setitem__(self, key, value):
+        self.data[key] = value
+
+    def __getitem__(self, key):
+        return self.data[key]
 
 def test_to_files():
     chat = textwrap.dedent(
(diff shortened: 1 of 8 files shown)


tests/test_chat_to_files.py
```diff
@@ -1,6 +1,21 @@
 import textwrap
 
 from gpt_engineer.core.chat_to_files import to_files
+from gpt_engineer.core.db import DB, DBs
+
+from typing import Union
+from pathlib import Path
+
+
+class StubDB(DB):
```
TheoMcCabe, line 10: Id appreciate some feedback here. I put this code in because i was getting permissions errors when running the unit tests on the build agent. My solution was to mock the file system calls. However my gut feeling from reading around is that this isn't the 'python' way of implementing this. Probably i should be using the patch library or somthing? I'm new to python though so im not sure - but i didnt want to spend an age on it and this works. If this is wrong a point in the right direction would be handy. Perhaps longer term we would want to look at a full file system mock for testing (https://pypi.org/project/pyfakefs/) ?

TheoMcCabe:
Oh i just noticed there is already a very similar PR out that addresses some of these issues! Maybe this is wasted effort

TheoMcCabe:
Closing as similar to https://github.com/AntonOsika/gpt-engineer/pull/749 which has more functionality and seems close to being merged

PR #766, opened 2023-10-04
Issue 718 refactor

captivus:
Resolves issue #718 and enhances documentation of each refactored module.

Tests running successfully on refactored codebase, with the exception of the misimplemented (missing) test for the AI module
image

PR #778, opened 2023-10-10
Bugfix - CLI - Resolves error "ModuleNotFoundError: No module named 'gpt_engineer.cli'"

captivus:
Resolves cli bug resulting from refactoring in #766.  Bug errors as follows:
```bash
ModuleNotFoundError: No module named 'gpt_engineer.cli'
```


captivus:
@ATheorell

PR #779, opened 2023-10-10
Adding tests for refactored CLI

captivus:
Given the bug surfaced and resolved with PR #778, added tests to cover this and related cases in the future.

Created 3 tests (and cleanup utility function) in test_install.py:
* Test that the package can be installed (locally) using `pip`
  * This is done in a virtual environment
  * The virtual environment persists throughout the rest of the tests
* Test that all modules can be imported
  * This tests only the imports of the core modules, as imports of the CLI modules is not recommended or supported (see issue #718)
* Tests that `gpt-engineer` command can be successfully run via the CLI
  * This tests the bug that was resolved in PR #778 and related bugs

After test execution, the virtual environment is removed.

I could use some guidance on how to separate this PR -- which I'm raising from the `refactored-cli-tests` branch which contains only the 3 CLI-related test addition commits -- from the other PR (#778) / branch that I created to solve the bug (branch `bugfix-cli-20231009`).   When I check the git log for the tests branch, I only see the 3 commits I want as part of this PR (which build on the other PR, admittedly).

image

I hope I've done this right!  Please share any comments or guidance.

### DIFF
--- a/gpt_engineer/__init__.py
+++ b/gpt_engineer/__init__.py
@@ -1,0 +1,8 @@
+# Adding convenience imports to the package
+from gpt_engineer.core import (
+    ai,
+    domain,
+    chat_to_files,
+    steps,
+    db,
+)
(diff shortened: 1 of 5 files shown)


captivus:
@ATheorell

PR #785, opened 2023-10-11
feat: add caching and benchmark

joschkabraun:
Hey, I opened a PR which allows you to quickly iterate on your app locally.

Adding the `init` statement will automatically use a local redis cache for any of your LLM requests (more [here](https://github.com/parea-ai/parea-sdk-py#debugging-chains--agents)). With that you won't need to wait for the slow & expensive API requests if you make any changes to your code/prompts and want to make sure everything works as expected.

This will also enable you to test gpt-engineer across many inputs at the same time via:

```bash
parea benchmark --func gpt_engineer/cli/main:main --csv_path benchmark-inputs.csv
```

The benchmark will create a CSV file with all the traces for you to debug.

I ran the benchmark with this CSV file:
```csv
project_path,model,temperature,steps_config,improve_mode,lite_mode,azure_endpoint,use_custom_preprompts,verbose
"projects/example1","gpt-4",0.1,"default","","","","",""
"projects/example2","gpt-4",0.5,"default","","","","",""
```

### DIFF
--- a/gpt_engineer/cli/main.py
+++ b/gpt_engineer/cli/main.py
@@ -32,6 +32,7 @@ from pathlib import Path
 import openai
 import typer
 from dotenv import load_dotenv
+from parea import init, RedisCache
 
 from gpt_engineer.core.ai import AI
 from gpt_engineer.core.db import DB, DBs, archive
@@ -40,6 +41,12 @@ from gpt_engineer.cli.collect import collect_learnings
 from gpt_engineer.cli.learning import collect_consent
 
 app = typer.Typer()  # creates a CLI app
+
+
+# set this to True to use caching of LLM calls and being able to benchmark function across many inputs in parallel
+IS_DEV = True
+if IS_DEV:
+    init(cache=RedisCache())
 
 
 def load_env_if_needed():
--- a/gpt_engineer/cli/main.py
+++ b/gpt_engineer/cli/main.py
@@ -30,9 +30,9 @@ import os
 from pathlib import Path
 
 import openai
+from parea import init, RedisCache
 import typer
 from dotenv import load_dotenv
-from parea import init, RedisCache
 
 from gpt_engineer.core.ai import AI
 from gpt_engineer.core.db import DB, DBs, archive
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ dependencies = [
   'python-dotenv >= 0.21.0',
   'langchain >=0.0.240',
   'agent-protocol==1.0.1',
+  'parea-ai >= 0.2.10',
 ]
 
 classifiers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
   'python-dotenv >= 0.21.0',
   'langchain >=0.0.240',
   'agent-protocol==1.0.1',
-  'parea-ai >= 0.2.10',
+  'parea-ai >= 0.2.11',
 ]
 
 classifiers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
   'python-dotenv >= 0.21.0',
   'langchain >=0.0.240',
   'agent-protocol==1.0.1',
-  'parea-ai >= 0.2.11',
+  'parea-ai >= 0.2.12',
 ]
 
 classifiers = [


joschkabraun:
CC: @AntonOsika @ATheorell  is that helpful for you guys?
E.g. the benchmarking can help you to assess the effect of this [recent pre-prompt change](https://github.com/AntonOsika/gpt-engineer/pull/795) in comparison to the old setup across many use cases, to make sure it's not breaking anything & has the desired effect.

TheoMcCabe:
So the purpose of this is to redirect traffic which would go to LLMs to a redis cache running on my local machine? So ican inspect the prompts created by my code changes?? What does the code receive back from the prompts - dummy text or nothing? 

Running an entire redis cache locally just to inspect the prompts produced by the code seems a little heavyweight. We can achieve similar  hings by patching the edges of the application and running in pytest ?? - or by using langchains FakeLLM ?

whats the benefit of this approach and why use something as powerful as a local redis cache for what seems to be just storing some text output by a python script?

ATheorell:
I concur with your doubts @TheoMcCabe . Closing for now.

joschkabraun:
@TheoMcCabe I should have been clearer:
> So the purpose of this is to redirect traffic which would go to LLMs to a redis cache running on my local machine?

No, it's checking if the exact same request was sent previously to the LLM. If so, it'll return the previous completion. If not it'll forward the request to the LLM, then store the completion in the cache and return the completion.

You get those benefits from adding the `init` statement:
- no wait time for redundant LLM calls
- logging & visualization of LLM calls

If you don't want to run a redis cache locally, you can simply create an API key for a free account on [Parea AI](www.parea.ai) and this enables you to do the same thing.

Note, if someone doesn't set their Parea API key or redis cache, it won't impact the experience of using `gpt-engineer`. The idea is to enhance your developer experience.

How does that sound?

TheoMcCabe:
Ah yes. This does sound useful now you've explained it! A good cost saver during development for sure.

TheoMcCabe:
We would need to work out how to turn caching on or off though. Perhaps with a flag on the main call ?

joschkabraun:
> We would need to work out how to turn caching on or off though. Perhaps with a flag on the main call ?

That's true. Happy to suggest an implementation, if that's of interest.

In case you don't want to use a (local) redis cache but use it through our hosted solution: we automatically check & read-in `PAREA_API_KEY`  from the OS environment variables and use our hosted cache if it exists. With that you don't need a DEV environment.

joschkabraun:
@ATheorell WDYT now with the additional context, should this PR be reopened?

captivus:
Thanks for the suggestion! We will discuss it further in an upcoming developers planning session.

joschkabraun:
Great! Out of curiosity when will that be?

joschkabraun:
@captivus any update from the developer planning session?

captivus:
Thanks for the PR, @joschkabraun!  We we want to incorporate this, and similar PRs, to our forthcoming GPT Engineer Cookbook.  We're keen to collect these types of use cases in the Cookbook as recipes to demonstrate use cases to the community.  We'll feedback shortly to ask for your contribution to the Cookbook.  Kindly standby!

joschkabraun:
Okay. I am not sure how this fits into the GPT Engineer Cookbook as the PR is meant as a developer tool while you improve/code on GPT Engineer. But, curious to see the GPT Engineer Cookbook and see how it fits into it. Happy to contribute!

captivus:
Help me understand your product better. As I understood it, you cache the prompt to, and response from, the LLM. In the example you suggest, as GPTE iterates on the target codebase, how does the cache assist the core functionality of GPTE?

joschkabraun:
You have the correct understanding of what the SDK does. The idea is to enhance the developer experience of someone who contributes to GPTE. At the same time, it doesn't impact the experience of anyone using GPTE. Does that make sense?

Concretely any developer of GPTE will get the following two benefit with the PR:

1. As you make changes to the GPTE code and try them out, you don't need to wait for redundant LLM calls
2. You can benchmark multiple different tasks or models at the same time (e.g. to understand does this only work for GPT-4 but also for GPT-3.5-turbo). The results of all the intermediate steps are then saved as a CSV such that you can investigate the individual steps if an execution failed.

Additionally, if you set your `PAREA_API_KEY` in the environment variables, you will get the cache out of the box without using a redis cache locally. This will also visualize those intermediate steps on a dashboard. You can use that to build a CI/CD pipeline & evaluate any code changes to GPTE more quantitatively. Note, you would get the `PARAE_API_KEY` for free for anyone who is developing GPTE as this is an open-source project.

How does that sound?

captivus:
I'll preface this by saying that I / we have no position on the value proposition of LLM caching products.

Caching LLM responses poses several problems to the gpt-engineer workflow due to the nondeterministic nature of these models.  If, as a developer, one expects a deterministic response to a specific prompt, one would save the response to the prompt in the due course of development (e.g. prompt saved in a variable, response saved in a variable). 

What am I missing in the proposed use case for gpt-engineer of the Parea product?

joschkabraun:
You wouldn't save the response to the prompt as the prompt is actually different every time if you use GPTE for a different purpose. Only the prompt template stays the same.

There are different use cases/benefits:

- if you fix a bug which is unrelated to the prompt and want to end-to-end test the changes along the way; by using the cache you won't have to wait for all the LLM calls but get their cached response
- If a developer iterates on a prompt which is used at later step (e.g. the [file format fix prompt](https://github.com/AntonOsika/gpt-engineer/blob/main/gpt_engineer/preprompts/file_format_fix)), then you could try out different formulations of it and see how it works out for the subsequent steps of GPTE without needing to regenerate the previous steps

Are those useful benefits?

@TheoMcCabe saw the benefit [here](https://github.com/AntonOsika/gpt-engineer/pull/785#issuecomment-1777436847)

PR #787, opened 2023-10-12
(Docs) Add support for Llama2, Claude, Palm, Cohere, Replicate (100+ LLMs)

krrishdholakia:
Hey @AntonOsika, 

Based on your feedback on this PR - https://github.com/AntonOsika/gpt-engineer/pull/660, 

My PR shows users how to call their custom LLM providers (bedrock, togetherai, huggingface tgi, replicate, ai21, cohere, ai21 etc.) by pointing the api base to a [local proxy](https://docs.litellm.ai/docs/proxy_server) they can use for experimentation with GPT-Engineer.

This makes **no code changes** to GPT-Engineer.

Happy to add additional tests/documentation if the initial PR looks good.

PR #788, opened 2023-10-12
Support for OpenAI Base API Integration

haseeb-heaven:
This PR introduces a feature mentioned [here]( https://github.com/AntonOsika/gpt-engineer/issues/710)

This feature introduces the capability for users to integrate their own **OpenAI Base API**. Currently, our system is reliant on the OpenAI's primary API Base. Given the associated costs and the availability of different models such as _GPT 3.5 and GPT 4_, it's imperative to provide an alternative. By allowing users to supply their own base API, we can potentially offer more cost-effective or even complimentary access to **GPT** models. This can be configured through a dedicated settings file, enabling users to utilize the **GPT Engineer** with their preferred version of **GPT**.

The command to Run GPT-Engineer using Custom base API.
```python
python -m gpt_engineer.cli.main projects/example 'gpt-4' --openai-api-base https://heaven-gpt.haseebmir.repl.co
```

**Current Configuration:**
1. The **API Base** is configured via the command line option `--openai-api-base`.
2. ~~At present, the streaming functionality is *deactivated*.~~ Updated the **Streaming response** now.
3. It's mandatory to set the **API Key** using the `.env` (Environment) file.

**Proposed Enhancements:**
1. ~~Implement a mechanism to modify the **API Base** for both streaming and non-streaming responses.~~ Updated in **latest PR**.
2. Allow the **API Key** to default to *Empty* when a specific base API is chosen. While this is feasible with LiteLLM, it presents challenges with LangChain and OpenAI. I am exploring more solutions to address this.
I have created this issue on LangChain about this here [LangChain OpenAI API Key Optional](https://github.com/langchain-ai/langchain/issues/11700) check it out.
3. Ask user to provide Authentication information for its own API server.
Details like `Is server Authenticated with API Key` and with this details we can make changes in codebase.

**_Test use cases for servers._**
**Custom** API servers can be of two types:

- **Authenticated**: These servers require users to authenticate themselves using a mechanism such as OAuth or API keys. This ensures that only authorized users can access the API.

- **Unauthenticated**: These servers do not require any authentication. This means that anyone can access the API, which can be a security risk.

I have only tested with **Unauthenticated** _server_ as i only have single server to test, and to make sure we cover all server we need to test the both server for our use case here.

### DIFF
--- a/gpt_engineer/cli/main.py
+++ b/gpt_engineer/cli/main.py
@@ -43,11 +43,15 @@ app = typer.Typer()  # creates a CLI app
 
 
 def load_env_if_needed(openai_api_base: str):
+    
     if openai_api_base:
-        openai.api_key = os.getenv("OPENAI_API_KEY")
-        # Should set to Empty key would work but its not working, Finding the solution.
+        openai.api_key = ""
+        # API Key is optional in case of Proxy servers (https://docs.litellm.ai/docs/proxy_server#replace-openai-base), 
+        # But in current situation its not working with empty key
+        # Finding another solutions here.
         return
-    elif os.getenv("OPENAI_API_KEY") is None:
+    
+    if os.getenv("OPENAI_API_KEY") is None:
         load_dotenv()
         if os.getenv("OPENAI_API_KEY") is None:
             # if there is no .env file, try to load from the current working directory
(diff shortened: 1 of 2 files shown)


ATheorell:
This runs for me.
@pbharrin and @AntonOsika, this goes beyond my skill level, would be great if you could give some input. 
I also suggest that, if we merge #787 , we should add this as an example of a server that can be run @krrishdholakia .
Comparing to #787 , maybe it would be enough to set OPENAI_API_BASE as an environment variable (no need for a new flag in that case)?

Issue #786, opened 2023-10-11
gpt-engineer doesn't respect the COLLECT_LEARNINGS_OPT_OUT=true env variable


peterjdolan:
## Policy and info
 - Maintainers will close issues that have been stale for 14 days if they contain relevant answers.
 - Adding the label "sweep" will automatically turn the issue into a coded pull request. Works best for mechanical tasks. More info/syntax at: https://docs.sweep.dev/

## Expected Behavior

When setting the environment variable COLLECT_LEARNINGS_OPT_OUT=true, no information should be transmitted back to the gpt-engineer developer.

## Current Behavior

Based on viewing the verbose execution output, it's clear that even with that environment variable set, information was transmitted back to the developer. On inspecting the consent methods, such as https://github.com/AntonOsika/gpt-engineer/blob/main/gpt_engineer/cli/learning.py#L172, it's clear that the environment variable is never referenced.

This is highly undesirable, considering that this is the mechanism for opting out of data collection described in the terms of use - https://github.com/AntonOsika/gpt-engineer/blob/main/TERMS_OF_USE.md.

## Failure Information

I've already transmitted too much information to the developer, and don't feel comfortable adding anything more.

ATheorell:
This is a problem with the TERMS_OR_USE (and test_collect.py) not being up to date.
Opting out of collecting learnings is now done with the .gpte_consent file. From learning.py:

```
def collect_consent() -> bool:
    """
    Check if the user has given consent to store their data.
    If not, ask for their consent.

    Returns
    -------
    bool
        True if the user has given consent, False otherwise.
    """
    consent_flag = Path(".gpte_consent")
    if consent_flag.exists():
        return consent_flag.read_text() == "true"

    if ask_if_can_store():
        consent_flag.write_text("true")
        print()
        print("(If you change your mind, delete the file .gpte_consent)")
        return True
    return False
```
Thus, what needs to be done is updating the TERMS_OF_USE.

ATheorell:
I made a suggestion in #797

captivus:
Fixed in PR #806 @ATheorell

PR #806, opened 2023-10-17
Fix:  Issue 786 - gpt-engineer doesn't respect the COLLECT_LEARNINGS_OPT_OUT=true env variable

captivus:
This PR fixes Issue #786.  It also revises, consolidates and simplifies the data collection consent workflow.  

* There is now only 1 function to check for user consent for data collection:  `cli/learning.py:check_collection_consent()`
* There is now only 1 function to interact with the user to obtain consent for data collection:  `cli/learning/ask_collection_consent()`
* Tests for each of these have been written in `tests/test_collection_consent.py`

As noted in the revised Terms of Use in PR #797, GPTE no longer uses the `COLLECT_LEARNINGS_OPT_OUT` environment variable; rather, uses the `.gpte_consent` file to manage user opt-in to data collection. This has been streamlined in this PR.

Updated preexisting test in `tests/test_collect.py` to reflect implementation of new data collection consent workflow (i.e. it solely tests the actual collection of data rather than checking for consent, as this is handled in the new workflow tests)

All tests successfully execute, with the exception of the (mis-implemented) `test_ai.py`.


captivus:
@ATheorell

gpt_engineer/cli/learning.py
```diff
@@ -146,76 +146,121 @@ def human_review_input() -> Review:
     )
 
 
-def check_consent() -> bool:
+# def check_collection_consent() -> bool:
+#     """
+#     Check if the user has given consent to store their data.
```
TheoMcCabe, line 151: I think we shouldn't commit commented code to main

TheoMcCabe:
Looks good to me - other than the commented code

TheoMcCabe (approved):

PR #813, opened 2023-10-20
Increase chat to files coverage

TheoMcCabe:


ATheorell:
Thanks a lot for the tests, but also for the cleanup. Everything seems to be working on my system

PR #817, opened 2023-10-23
Update ai.py

mfeyx:
I added the ability to set OPENAI_API_VERSION from the .env file. There is a newer API version available in the Azure OpenAI Workspace ("2023-07-01-preview"), and it would be nice to use it instead of the May version.


TheoMcCabe:
looks good thanks

TheoMcCabe:
looks good thanks

PR #828, opened 2023-10-28
Vector store - initial unrefined implementation 

TheoMcCabe:
This is a first pass at using a vector store to automatically retrieve files from a code repository based on their relevance to the prompt.

- Uses llama index for the vector store abstraction 
- Implements a code splitter (with most of the code taken from llama index) which depends on tree sitter library 
- Creates a new step set called VECTOR_IMPROVE to run this improve which is run with the -vi argument 
- Finds 2 relevant snippets then feeds in the entire file for those snippets 
- Also provides a list of all code files in the repository to provide wider context 

Theres a lot more work to do in this area, but I think the scope of this work is enough to merge on its own pending review etc. It works for me for at least small use cases like snake, so seems be a good place to hand over to others to have  a play around with? 

Some ideas of what to do next: 
- The whole vector store piece needs lots of refining and is largely untested on large repositories etc. An obvious limitation is that its currently hard coded to retrieve 2 snippets, and return the files those snippets are in. So a maximum of 2 files will be fed to the LLM. Possibly we would want to load as many files as possible within the token limit?
- Other improvements we might want to look into would be an improved ranking of snippets based on connectedness ? connectness of functions is currently calculated in aider
- An improved high level context. Aider uses tree sitter to sumarise each file into methods and classes, and analyses the connectness and then returned a summaries tree representation of the most important parts of the code. 
- refine the retrieval algorithms - to work better for code and possibly to include additional meta data in the query 
- expand the vector store to include non code files


TheoMcCabe:
@ATheorell  pointed out ive been committing with the wrong email address  - oops!! Just created a new PR with the author fixed ... i wont retrospectively attempt to change main ... wouldnt want to cause issues with changed commit hashes

PR #840, opened 2023-11-07
Refactor

ATheorell:
#UPDATED VERSION OF THE STATE OF THE REFACTOR

So, here comes the refactor! Note that this PR goes to a new branch called refactor and NOT to main, so we can merge it and collaborate on it without side effects and work on the remaining problems asynchronously. The reason why I put the refactor up in this half state is so that more feedback can be integrated early on.

Short explanation of whats going on here:

    The most important thing is the creation of an agent class that should make it easy to create diverse application. At this point, only init, which creates code from scratch is implemented. The signature of improve may change to take a "Code" as input.
    Agent is now the main entrypoint into the code and typically, one agent will be written per application. The Agent gets a configuration upon instantiation and handles step selection. 
    The directory "default" inside core provides minimal implementations of the abstract interfaces, so that its possible to have such a default setup without importing from applications (having mutual dependencies between core and applications seems wrong).
    Maybe the biggest shift in philosophy is that dbs no longer rule the world. In the latest take, the Agents do not have dbs nor access to a path to the working directory. Instead, (potential but not mandatory) IO is handled by a Memory, an ExecutionEnv and a VersionManager. The VersionManager lies outside the scope of the Agent. 
    The Version manager is yet to be implemented and takes Code instances and creates snapshots. The default Version manager should be git.
    Without universal FileRepositories, there is no longer a unified interface for all steps. To keep in memory Code isntances as the main vehicle of code, while at the same time keeping the steps stateless, it is sometimes necessary to pass a Memory or ExecutionEnv instance.

What works and doesn't work currently:

    The cli application with the default flags works, probably nothing else.
    improve is probably the largest feature yet untouched
    VersionManager is not implemented
    The tests don't work
    Much is undocumented
    Some more intelligent way to handle the preprompts is desired.
    The "learning" part of the cli application is more or less untouched and still uses code from the "legacy" steps.
    I somehow broke the token counting, by not saving logs anymore. Not sure where I messed that up or where that should go in.
    Initialization in a repo with existing code is yet to be implemented.

Any input on the architecture is more than welcome ASAP!

Birds eye view:
```
├── applications
│   ├── cli
│   │   ├── collect.py
│   │   ├── file_selector.py
│   │   ├── __init__.py
│   │   ├── learning.py
│   │   ├── main.py
│   ├── __init__.py
├── core
│   ├── ai.py
│   ├── base_agent.py
│   ├── base_execution_env.py
│   ├── base_repository.py
│   ├── base_version_manager.py
│   ├── chat_to_files.py
│   ├── code.py
│   ├── default
│   │   ├── git_version_manager.py
│   │   ├── __init__.py
│   │   ├── lean_agent.py
│   │   ├── on_disk_execution_env.py
│   │   ├── on_disk_repository.py
│   │   ├── paths.py
│   │   └── steps.py
│   ├── domain.py
│   ├── __init__.py
│   └── token_usage.py
├── data
│   ├── code_vector_repository.py
│   ├── document_chunker.py
│   ├── __init__.py
│   └── supported_languages.py
├── __init__.py
├── legacy
│   ├── __init__.py
│   └── steps.py
├── preprompts

...
```


gpt_engineer/core/step_bundle_interface.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+from gpt_engineer.core.code import Code
+from gpt_engineer.core.ai import AI
+
+
+class StepBundleInterface(ABC):
```
TheoMcCabe, line 6: I think BaseStepsBundle is a more pythonic name

gpt_engineer/cli/main.py
```diff
@@ -1,198 +0,0 @@
-"""
```
TheoMcCabe, line 1: is this the only file deleted from /cli ?

gpt_engineer/core/code.py
```diff
@@ -0,0 +1,13 @@
+from typing import TypeVar, Dict
+
+class Code(dict):
```
TheoMcCabe, line 3: Is this a code file or a snippet? or is it supposed to be either?

gpt_engineer/core/default/steps.py
```diff
@@ -0,0 +1,257 @@
...
+from langchain.schema import HumanMessage, SystemMessage
+
+import inspect
+import os
+import re
+from termcolor import colored
+import subprocess
+
+#TODO: THIS NEEDS A BETTER SOLUTION
+PREPROMPTS_PATH = os.path.join("gpt_engineer", "preprompts")
```
TheoMcCabe, line 15: Shouldnt pre prompts just be injected as constant strings from a python module

gpt_engineer/applications/cli/main.py
```diff
@@ -0,0 +1,201 @@
...
+    if input_repo.get("prompt"):
+        return input_repo.get("prompt")
+
+    input_repo["prompt"] = input(
+        "\nWhat application do you want gpt-engineer to generate?\n"
+    )
+    return input_repo.get("prompt")
+
+
+# def preprompts_path(use_custom_preprompts: bool, input_path: Path = None) -> Path:
```
TheoMcCabe, line 66: remove comments?

gpt_engineer/cli/main.py
TheoMcCabe: Never mind i can see the other files are marked as moved

tests/test_chat_to_files.py
```diff
@@ -1,7 +1,7 @@
 import textwrap
```
TheoMcCabe, line 1: I think we should arrange tests colder into core, cli, applications folders to match the source code

TheoMcCabe (commented):
looks good just a few comments

gpt_engineer/core/code.py
ATheorell: This is supposed to be a dictionary[filepath] = file_content, potentially containing the whole codebase.

gpt_engineer/core/code.py
```diff
@@ -0,0 +1,13 @@
...
+
+    def __setitem__(self, key, value):
+        if not isinstance(key, str):
+            raise TypeError("Keys must be strings")
+        if not isinstance(value, str):
+            raise TypeError("Values must be integers")
+        super().__setitem__(key, value)
```
AntonOsika, line 10: Nitpick:

Does this work?:)

```suggestion
        super()[key] = value
```

gpt_engineer/core/step_bundle_interface.py
```diff
@@ -0,0 +1,14 @@
...
+
+    @abstractmethod
+    def gen_code(self, ai: AI, prompt: str) -> Code:
```
AntonOsika, line 9:
```suggestion
    def init(self, ai: AI, prompt: str) -> Code:
```

gpt_engineer/core/step_bundle_interface.py
```diff
@@ -0,0 +1,14 @@
...
+        pass
+
+    @abstractmethod
+    def improve_code(self, ai: AI, prompt: str) -> Code:
```
AntonOsika, line 13:
```suggestion
    def improve(self, ai: AI, prompt: str) -> Code:
```

AntonOsika: Shouldn't this accept Code as an argument?

(Also, maybe we should give access to "memory"?)

gpt_engineer/core/default/steps.py
AntonOsika: Just want to point out the history here, with the ambition that it should be easy to switch out the preprompts by the _user_ of gpt-engineer.

(Or even use something like scale.ai spellbook in the future)

I'm all for simplifying now and add that back later.

gpt_engineer/applications/cli/cli_step_bundle.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, workspace_path: str):
+        self.workspace_path = workspace_path
+        self.memory = FileRepository(memory_path(workspace_path))
+
+
+    def gen_code(self, ai: AI, prompt: str) -> Code:
+        code = gen_code(self.workspace_path, ai, prompt, self.memory)
+        #TODO: evaluate whether it makes more sense to send the code than the memory to gen_entrypoint
+        entrypoint = gen_entrypoint(self.workspace_path, ai, self.memory)
```
AntonOsika, line 20: Code is more specific so should def send that one imo

gpt_engineer/core/agent.py
```diff
@@ -0,0 +1,64 @@
...
+            It also snapshots the improved code using the version manager.
+
+            Parameters:
+                prompt (str): A string prompt that guides the code improvement process.
+
+            Returns:
+                Code: An instance of the `Code` class containing the improved code.
+    """
+
+    def __init__(self, path: str, version_manager: VersionManagerInterface = None, step_bundle: StepBundleInterface = None, ai: AI = None):
```
AntonOsika, line 49: Not super important, but I think it's a bit many input fields here!

Would much prefer to have init without path like:

```
def __init__(self, version_manager: VersionManagerInterface, step_bundle: StepBundleInterface, ai: AI = None):
        self.path = path
        self.version_manager = version_manager
        self.step_bundle = step_bundle
        self.ai = ai or AI()```
        
And then a separate path constructor like this:

```
@classmethod
def from_path(cls, path: str, ai: AI = None):
    return self(
       version_manager = VersionManager(self.path)
       step_bundle = LeanStepBundle(self.path)
       ai = ai or AI()
    )

```

Comment (AntonOsika) [reply gpt_engineer/core/agent.py:49]: so that one can do:

```
Agent.from_path(path_to_snake_project, ai=AI(...))
```

File: gpt_engineer/core/code.py (hunk 3 above)
Comment (AntonOsika) [reply gpt_engineer/core/code.py:3]:
```suggestion
class Code(MutableMapping[str | Path, str]):
```

gpt_engineer/core/code.py
```diff
@@ -0,0 +1,13 @@
+from typing import TypeVar, Dict
+
+class Code(dict):
+
+    def __setitem__(self, key, value):
+        if not isinstance(key, str):
+            raise TypeError("Keys must be strings")
+        if not isinstance(value, str):
+            raise TypeError("Values must be integers")
```
AntonOsika, line 9:
```suggestion
            raise TypeError("Values must be strings")
```

AntonOsika:
Should we maybe rename data -> rag? (or put it in tools)

And if we keep file_repository, shouldn't that be in core?

gpt_engineer/core/agent.py
```diff
@@ -0,0 +1,64 @@
...
+        self.path = path
+        self.version_manager = version_manager or VersionManager(self.path)
+        self.step_bundle = step_bundle or LeanStepBundle(self.path)
+        self.ai = ai or AI()
+
+    def init(self, prompt: str) -> Code:
+        code = self.step_bundle.gen_code(self.ai, prompt)
+        version_hash = self.version_manager.snapshot(code)
```
AntonOsika, line 57: Let's move the version_manager outside the agent?

Should be easy to just do versioning on the implementation side!

gpt_engineer/applications/cli/cli_step_bundle.py
```diff
@@ -0,0 +1,28 @@
...
+from gpt_engineer.core.default.steps import gen_code, gen_entrypoint, execute_entrypoint
+from gpt_engineer.core.default.paths import memory_path
+from gpt_engineer.applications.cli.learning import human_review
+
+
+class CliStepBundle(StepBundleInterface):
+
+    def __init__(self, workspace_path: str):
+        self.workspace_path = workspace_path
+        self.memory = FileRepository(memory_path(workspace_path))
```
AntonOsika, line 14: This looks a bit weird to me.

Why does stepbundleinterface know about the path?

Why is memory always on disk? While Code is in memory and version management is abstract..

Would be much more reasonable to pass in the memory object!

gpt_engineer/core/default/steps.py
```diff
@@ -0,0 +1,257 @@
...
+    set up and functional. Ensure these prerequisites are in place before invoking `simple_gen`.
+    """
+    db = FileRepository(PREPROMPTS_PATH)
+    messages = ai.start(setup_sys_prompt(db), prompt, step_name=curr_fn())
+    chat = messages[-1].content.strip()
+    memory[CODE_GEN_LOG_FILE] = chat
+    files = parse_chat(chat)
+    workspace = FileRepository(workspace_path)
+    for file_name, file_content in files:
+        workspace[file_name] = file_content
```
AntonOsika, line 83: why do we write to disk here?

I thought a big idea behind this refactor was that we abstract the IO and only modify a Code object in memory.

AntonOsika: image

gpt_engineer/core/default/steps.py
```diff
@@ -0,0 +1,257 @@
...
+        step_name=curr_fn(),
+    )
+    print()
+
+    regex = r"```\S*\n(.+?)```"
+    matches = re.finditer(regex, messages[-1].content.strip(), re.DOTALL)
+    entrypoint_code = Code({ENTRYPOINT_FILE: "\n".join(match.group(1) for match in matches)})
+    # write entrypoint code to file
+    for key, val in entrypoint_code.items():
+        FileRepository(workspace_path)[key] = val
```
AntonOsika, line 138: Same here, I don't see why we write to disk here

gpt_engineer/core/step_bundle_interface.py
ATheorell: Yes, the signature of this will change and I'm thinking about making two improve classes, one that takes one Code (standard) and one that takes 2 Code (for improving a certain part). The second part will be crucial for writing larger programs.

gpt_engineer/applications/cli/cli_step_bundle.py
ATheorell: Feels neater! The open question to me is whether there is additional information in the non-code parts of the output which makes it easier for the LLM to understand what the entrypoint should run. I'll change it for now and keep this in mind if it starts under performing.

gpt_engineer/applications/cli/cli_step_bundle.py
ATheorell: Regarding this and the following comments:

I will make two abstract classes for memory and execution_environment and lean implementations of these will hold a path and keep the flow working roughly like it works now.

gpt_engineer/core/code.py
TheoMcCabe: Probably just different philosophies here - but i find 'Code' to be an ambiguous name which will makes our code harder to read. If its a key value store of files which make up a code base then iCodeFileStore, CodeFileRepository or just CodeFiles might make more sense? If it can include non code files maybe FileStore or FileRepository. 

As code can be any of an expression, a line, a snippet, a file or a project, a repository, and many other things in between -  its unclear what the objects in the mutuable mapping represent if you just call it Code

TheoMcCabe:
> Should we maybe rename data -> rag? (or put it in tools)
> 
> And if we keep file_repository, shouldn't that be in core?

My thinking with the data folder, is to separate logic to do with read and write and persistence from the core logic, which i think is the same lines your thinking too

If core is purely in memory operations - you would need a separate package to read and write from disk. This could also cover sql databases, vector stores , etc. so maybe in an application you would either read using data libraries, run core logic and then persist  using the data libraries, or inject concrete data handlers from the data libraries into the core classes which would read and write it as part of the core logic.

gpt_engineer/core/default/steps.py
TheoMcCabe: See my comment below - but if we make a rule that core library can not reference the data library then you will guarantee nothing will get written to disk by the core library

gpt_engineer/core/code.py
ATheorell: No it doesn't! Didn't take time to dig into why

PR #864, opened 2023-11-23
Refactor

ATheorell:
And here it comes! The request to merge the refactored gpt-engineer into main.

Main changes:

- Agent classes working as main interface to applications 
- Most operations now in-memory (though executions are still on disk)
- No longer a unified Step interface
- New file structure
- Much better test coverage
- and probably much more that I fail to think of now!

Reviews highly appreciated!


gpt_engineer/applications/cli/collect.py
```diff
@@ -120,15 +115,51 @@ def collect_learnings(
...
+    - Ensure that the database's memory has enough space or is set up correctly to store the serialized review tools.
     """
-    with open(steps.__file__, "r") as f:
-        content = f.read()
-        return hashlib.sha256(content.encode("utf-8")).hexdigest()
+
+    """Collects and stores human review of the code"""
+    review = human_review_input()
+    collect_learnings(prompt, model, temperature, config, memory, review)
+    return []
```
UmerHA, line 165: Why return a constant (here: the empty list)?

gpt_engineer/applications/cli/main.py
```diff
@@ -0,0 +1,216 @@
...
+        "-i",
+        help="Improve code from existing project.",
+    ),
+    vector_improve_mode: bool = typer.Option(
+        False,
+        "--vector-improve",
+        "-vi",
+        help="Improve code from existing project using vector store.",
+    ),
+    lite_mode: bool = typer.Option(
```
UmerHA, line 109: From a user's perspective, what's the difference between "improving code" and "improving code using a vector store"?

I would expect the user intent to be "improve my code" and using a vector db or not is a technical implementation detail that should not be exposed to the user?

gpt_engineer/applications/cli/main.py
```diff
@@ -0,0 +1,216 @@
...
+app = typer.Typer()  # creates a CLI app
+
+
+def load_env_if_needed():
+    if os.getenv("OPENAI_API_KEY") is None:
+        load_dotenv()
+    if os.getenv("OPENAI_API_KEY") is None:
+        # if there is no .env file, try to load from the current working directory
+        load_dotenv(dotenv_path=os.path.join(os.getcwd(), ".env"))
+    openai.api_key = os.getenv("OPENAI_API_KEY")
```
UmerHA, line 61: what if there is still not api key? we should notify the user.

gpt_engineer/applications/cli/main.py
```diff
@@ -0,0 +1,216 @@
...
+        code_gen_fn = lite_gen
+    else:
+        code_gen_fn = gen_code
+    # configure execution function
+    if self_heal_mode:
+        execution_fn = self_heal
+    else:
+        execution_fn = execute_entrypoint
+
+    if --vector_improve_mode:
```
UmerHA, line 180: shoud be "vector_improve_mode" not "--vector_improve_mode" unless im missing sth

gpt_engineer/applications/cli/main.py
```diff
@@ -0,0 +1,216 @@
...
+    agent = CliAgent.with_default_config(
+        project_path,
+        ai=ai,
+        code_gen_fn=code_gen_fn,
+        execute_entrypoint_fn=execution_fn,
+        improve_fn=improve_fn,
+        preprompts_holder=preprompts_holder,
+    )
+    if improve_mode:
+        if --vector_improve_mode:
```
UmerHA, line 196: shoud be "vector_improve_mode" not "--vector_improve_mode" unless im missing sth

gpt_engineer/core/default/git_version_manager.py
```diff
@@ -0,0 +1,28 @@
...
+        self.path = path
+
+    def snapshot(self, code: Code) -> str:
+        workspace = OnDiskRepository(self.path)
+        for file_name, file_content in code.items():
+            workspace[file_name] = file_content
+        # Typically save to file and/or git and return a hash
+
+        # Todo: Implement push to git!
+        return ""
```
UmerHA, line 28: dont forget to implement this ;)

gpt_engineer/core/default/steps.py
```diff
@@ -0,0 +1,183 @@
...
+
+def gen_code(
+    ai: AI, prompt: str, memory: BaseRepository, preprompts_holder: PrepromptsHolder
+) -> Code:
+    preprompts = preprompts_holder.get_preprompts()
+    messages = ai.start(setup_sys_prompt(preprompts), prompt, step_name=curr_fn())
+    chat = messages[-1].content.strip()
+    memory[CODE_GEN_LOG_FILE] = chat
+    files = parse_chat(chat)
+    code = Code({key: val for key, val in files})
```
UmerHA, line 50: minor point: this is identical to `Code(**files)`

gpt_engineer/core/default/steps.py
```diff
@@ -0,0 +1,183 @@
...
+    messages = [
+        SystemMessage(content=setup_sys_prompt_existing_code(preprompts)),
+    ]
+    # Add files as input
+    messages.append(HumanMessage(content=f"{code.to_chat()}"))
+    messages.append(HumanMessage(content=f"Request: {prompt}"))
+    problems = [""]
+    # check edit correctness
+    edit_refinements = 0
+    while len(problems) > 0 and edit_refinements <= MAX_EDIT_REFINEMENT_STEPS:
```
UmerHA, line 168: make sense!

UmerHA (commented):
hey, have reviewed

gpt_engineer/applications/cli/collect.py
ATheorell: Removed

gpt_engineer/applications/cli/main.py
ATheorell: As long as its not explained to the user what the difference is, I agree 100%. 

Concretely, improve requires the user to provide which files to change, whereas vector improve improve picks code chunks to improve automatically from the whole code base. Probably for a UX perspective, maybe the top-level names for should be something like improve_manual_file_selection and improve_automatic_file_selection? Quite bulky... What is your opinion @TheoMcCabe  ?

Additional info here is that vector improve should be seen more as a first implementation and is not a mature default workflow.

gpt_engineer/applications/cli/main.py
ATheorell: I'm not really sure how you mean. If no key is provided, an openai.error.AuthenticationError is raised with a failure message. We can of course catch this, etc, but is this not informative enough?

gpt_engineer/core/default/git_version_manager.py
ATheorell: https://github.com/AntonOsika/gpt-engineer/issues/868

gpt_engineer/applications/cli/main.py
UmerHA: Or `improve_all` and `improve_files`

gpt_engineer/core/default/steps.py
ATheorell: Hmm, I don't get that to work for files, which is List[Tuple[str, str]] .

>>> a = [("z","x"), ("y","w")]
>>> b = dict(**a)
  File "<stdin>", line 1, in <module>
TypeError: dict() argument after ** must be a mapping, not list

Note that Code inherits dict.

gpt_engineer/applications/cli/main.py
UmerHA: I would catch that case and throw an Error with a more descriptive error message with a suggested action. This is optional though.

.langchain.db
ErikBjare, line 1: Was this added intentionally?

gpt_engineer/applications/cli/__init__.py
```diff
@@ -7,9 +7,9 @@
 
 Modules:
     - main: The primary CLI module for GPT Engineer.
-    - collect: Collect send learning data for analysis and improvement.
+    - collect: Collect send learning tools for analysis and improvement.
     - file_selector: Selecting files using GUI and terminal-based file explorer.
-    - learning: Tools and data structures for data collection.
+    - learning: Tools and tools structures for tools collection.
```
ErikBjare, line 12: Looks like a search-replace gone wrong?

gpt_engineer/applications/cli/collect.py
```diff
@@ -25,39 +25,31 @@
...
 
 
 def send_learning(learning: Learning):
     """
-    Send the learning data to RudderStack for analysis.
+    Send the learning tools to RudderStack for analysis.
 
     Note:
-    This function is only called if consent is given to share data.
+    This function is only called if consent is given to share tools.
```
ErikBjare, line 44: Lots of these s/data/tools that seems to be a mistake, might want to grep the diff for `tools` and go through them all.

gpt_engineer/applications/cli/cli_agent.py
```diff
@@ -0,0 +1,127 @@
...
+
+        improve(self, prompt: str) -> Code:
+            Improves an existing piece of code using the AI and step bundle based on the provided prompt.
+            It also snapshots the improved code using the version manager.
+
+            Parameters:
+                prompt (str): A string prompt that guides the code improvement process.
+
+            Returns:
+                Code: An instance of the `Code` class containing the improved code.
```
ErikBjare, line 64: I'm curious: why aren't these generated with things like [autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html)?

The project has a lot of docstrings which seem a pain to manually maintain and keep up-to-date, which autodoc would simplify massively (at the cost of not getting it exactly how you want it).

Reminds me of [this video](https://www.youtube.com/watch?v=k5DJxz6_xMA)

gpt_engineer/applications/cli/__init__.py
ATheorell: agreed

gpt_engineer/applications/cli/cli_agent.py
ATheorell: Would be amazing if you could have a shot at this once we merge.

gpt_engineer/applications/cli/cli_agent.py
```diff
@@ -0,0 +1,127 @@
...
+        self.memory = memory
+        self.execution_env = execution_env
+        self.ai = ai or AI()
+        self.code_gen_fn = code_gen_fn
+        self.process_code_fn = process_code_fn
+        self.improve_fn = improve_fn
+        self.preprompts_holder = preprompts_holder or PrepromptsHolder(PREPROMPTS_PATH)
+
+    @classmethod
+    def with_default_config(
```
ErikBjare, line 86: This doesn't seem to do anything different than `__init__`, can be removed.

PR #873, opened 2023-11-29
Fix leftover ai.fuser() call on self_heal

scodeman:
There is a leftover call to `ai.fuser()` on the `self_heal` step that prevent using the step with error 
```
  File "/usr/local/bin/gpt-engineer", line 8, in <module>
    sys.exit(app())
  File "/app/gpt_engineer/cli/main.py", line 186, in main
    messages = step(ai, fileRepositories)
  File "/app/gpt_engineer/core/steps.py", line 723, in self_heal
    messages.append(ai.fuser(get_platform_info()))  # add in OS and Py version
AttributeError: 'AI' object has no attribute 'fuser'
```
fix is making the change to use `HumanMessage` function directly like on other steps


ATheorell:
Thanks for the fix!

PR #890, opened 2023-12-06
Remove excess whitespace in files included in context.

Gullesnuffs:
This PR fixes a bug causing files included in the LLM context to be formatted in a weird way.

Using the current code, a file will be formatted like this:
```
    main.jsx
    ```
    import React from "react";
import ReactDOM from "react-dom/client";
import App from "./App.jsx";
import "./index.css";

ReactDOM.createRoot(document.getElementById("root")).render(
  <React.StrictMode>
    <App />
  </React.StrictMode>
);

    ```
```

Note the extra indentation at the first three lines and the last line, which I think may be confusing for the LLM (and even if it isn't confusing, it wastes a handful of tokens).

### DIFF
--- a/gpt_engineer/core/chat_to_files.py
+++ b/gpt_engineer/core/chat_to_files.py
@@ -175,11 +175,11 @@ def format_file_to_input(file_name: str, file_content: str) -> str:
         The formatted file string.
     """
     file_str = f"""
-    {file_name}
-    ```
-    {file_content}
-    ```
-    """
+{file_name}
+```
+{file_content}
+```
+"""
     return file_str
 
 


ATheorell:
Good point here @Gullesnuffs . In the (huge) refactor that is soon to be merged, this code has moved into the class FilesDict and I have actually already done exactly this edit independently: https://github.com/AntonOsika/gpt-engineer/blob/refactor-poetry/gpt_engineer/core/files_dict.py (feel free to confirm its the same). It got stuck in a new test I made somehow. Hopefully the refactor will be merged later today or tomorrow!

Gullesnuffs:
Yes, that looks the same edit. Nice! :)

PR #920, opened 2023-12-20
build(deps): moved dev dependencies to dev group

ErikBjare:
Found it weird that `pytest`, `ruff`, etc was in the main dependency group. This fixes that.


pyproject.toml
```diff
@@ -10,11 +10,6 @@ repository = "https://github.com/AntonOsika/gpt-engineer"
 documentation = "https://gpt-engineer.readthedocs.io/en/latest/"
 classifiers = [
   "Development Status :: 4 - Beta",
-  "Programming Language :: Python :: 3.8",
-  "Programming Language :: Python :: 3.9",
-  "Programming Language :: Python :: 3.10",
-  "Programming Language :: Python :: 3.11",
-  "License :: OSI Approved :: MIT License",
```
ErikBjare, line 17: These are set by default by poetry: https://python-poetry.org/docs/pyproject/#classifiers

PR #922, opened 2023-12-21
fix CLI parameter at doc

nnashiki:
"Thank you for providing such a great product. I found a typo in the doc while using it. I have confirmed that it works when using the CLI parameter --use-custom-preprompts. The basis for the correction is in the following section:

gpt_engineer/applications/cli/main.py
``` 
    use_custom_preprompts: bool = typer.Option(
        False,
        "--use-custom-preprompts",
        help="""Use your project's custom preprompts instead of the default ones.
          Copies all original preprompts to the project's workspace if they don't exist there.""",
    ),
```


ErikBjare:
Thanks! :)

Issue #928, opened 2023-12-22
KeyError in apply_edits breaking improve mode


baldmanwithbeard:
I am running improve mode, creating c# and xaml. GPT Engineer is attempting to make updates to a xaml user control (here renamed to be "myExistingUserControl.xaml") and running into an issue where the filepath is invalid.

```These edits will ensure that the code changes are in the correct format and can be found in the code.Traceback (most recent call last):

  File "<frozen runpy>", line 198, in _run_module_as_main

  File "<frozen runpy>", line 88, in _run_code

  File "C:\Users\asdf\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\Scripts\gpte.exe\__main__.py", line 7, in <module>
    sys.exit(app())
             ^^^^^

  File "C:\Users\asdf\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\site-packages\gpt_engineer\applications\cli\main.py", line 194, in main
    files_dict = agent.improve(files_dict, prompt)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users\asdf\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\site-packages\gpt_engineer\applications\cli\cli_agent.py", line 131, in improve
    files_dict = self.improve_fn(
                 ^^^^^^^^^^^^^^^^

  File "C:\Users\asdf\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\site-packages\gpt_engineer\core\default\steps.py", line 182, in improve
    overwrite_code_with_edits(chat, files_dict)

  File "C:\Users\asdf\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\site-packages\gpt_engineer\core\chat_to_files.py", line 97, in overwrite_code_with_edits
    apply_edits(edits, files_dict)

[...]

Issue #869, opened 2023-11-25
Better edit syntax for the improve command


ATheorell:
## Feature description
gpt-engineer supports improving existing code with natural language prompts using the --improve flag. On a technical level, gpt-engineer [improves](https://github.com/AntonOsika/gpt-engineer/blob/refactor/gpt_engineer/core/default/steps.py) code by sending the improve prompt and the code to the LLM and asking the LLM to provide improvements in the form of edit blocks on the format of the given example:

```python
some/dir/example_1.py
<<<<<<< HEAD
    def mul(a,b)
=======
    def add(a,b):
>>>>>>> updated
```
This idea is largely borrowed from [aider](https://github.com/paul-gauthier/aider/blob/main/aider/coders/editblock_coder.py) .
However, this approach is rather error prone, since it either requires the LLM to output the HEAD part of the edit block to be identical to a part of the code, which quite often is not the case, or that some intelligent heuristics handle the case when no exact match is found. Problems with the current implementation are reported in  #721 #814 #841. 

An alternative way to make edits is to prompt the LLM to provide edits in the classic diff syntax
```
28 -    workspace = FileRepository(eval_ob["project_root"])
28 +    workspace = OnDiskRepository(eval_ob["project_root"])
```
This has the advantage that the file name + the line number uniquely defines where to put in the edit, enabling edits, even if the existing code is not reproduced perfectly. Line numbers are currently not stored and not provided to the LLM. The easiest way to do this is probably to implement a method that equips a code object with line numbers in the the [Code](https://github.com/AntonOsika/gpt-engineer/blob/refactor/gpt_engineer/core/code.py) class and update the [to_chat](https://github.com/AntonOsika/gpt-engineer/blob/refactor/gpt_engineer/core/code.py) method to provide the line numbers, exactly the way it looks in the diff syntax. The same line numbers should then be reused when parsing edits from the LLM back into the code.

Of course, to make the LLM understand what is going on, it is also necessary to modify the corresponding [preprompt](https://github.com/AntonOsika/gpt-engineer/blob/refactor/gpt_engineer/preprompts/improve).

## Motivation/Application
Making the --improve workflow 10x more reliable!

PR #930, opened 2023-12-23
potential fix for some/dir/ bug #928

ATheorell:
The problem in bug #928 is that "some/dir/" is used as a placeholder to indicate that a filename should be represented by a relative path. The placeholder is introduced in the preprompt improve. Occasionally, the LLM misunderstands this intention and appends "some/dir/" explicitly to file names. This fix adds a clarification to the LLM the "some/dir/" is a placeholder.


ErikBjare:
I think the `some/dir/` could perhaps be omitted entirely, and instead mention that it should output a file path, not file name.

Also not sure if this is the right place/time, but I've had great success by writing [save](https://github.com/ErikBjare/gptme/blob/master/gptme/tools/save.py) codeblocks like this:

    ```hello.py
    print("hello world")
    ```

And similarly for [patch](https://github.com/ErikBjare/gptme/blob/master/gptme/tools/patch.py), simply:

    ```patch hello.py
    <<<<<<< ORIGINAL
    print("Hello world")
    =======
    name = input("What is your name? ")
    print(f"hello {name}")
    >>>>>>> UPDATED
    ```

I think this might be a superior format overall, but not sure what your experiences with the current one is.

ATheorell:
Thanks for input @ErikBjare !
I'll have test removing the placeholder entirely. Regarding the syntax you propose, the risk I see is that the ORIGINAL part of the block may occur multiple times in the file, and we will by default replace all, even if we only want to modify one. The current (cumbersome) syntax tries to avoid this by always including a function name (unique by design).

I actually made feature request for a third idea https://github.com/gpt-engineer-org/gpt-engineer/issues/869
...and really, I have no idea what is the best way to go -> we need a proper benchmarking flow -> which is under construction! 

For now, as I said, I will try just removing the placeholder.

ATheorell:
This should at least partially address the concern in #928, by removing the part of the prompt that misleads the LLM to append some/dir to paths 

I have also added an additional test for the case that we are editing a new file (works) and overwriting an existing file (works). This, as I understand should address #952 , which I will close. Correct me if I'm misunderstanding something here @Wheaties466 .

ATheorell:
Great! I merged to main

PR #937, opened 2023-12-27
Add tree-style file selection feature

similato87:
Implement Tree-Style File Selection

Introduced a tree-style file selection method, allowing users to easily select files via a .toml configuration file. This update includes:

- Removal of the graphical file selection dependency.
- Implementation of a tree representation for file selection.
- Direct editing of file selection through the user's default text editor.
- Enhanced feedback by displaying selected files in a tree-style format before proceeding.

This update enhances usability for terminal-based interactions and streamlines the file selection process. Below is the gif for the feature:

feature

PR #948, opened 2024-01-02
New file selection fixed tests and rel path problem

ATheorell:
This is one commit ahead of #937

My fixes to the tests. Feel free to integrate these change in your original PR #937 and we can continue there @similato87


ATheorell:
Oops uploaded wrong branch! one more try

PR #949, opened 2024-01-02
Fixes to tests and paths rel to cwd/project dir

ATheorell:
This is one commit ahead of https://github.com/gpt-engineer-org/gpt-engineer/pull/937

My fixes to the tests. Feel free to integrate these change in your original PR https://github.com/gpt-engineer-org/gpt-engineer/pull/937 and we can continue there @similato87


similato87:
Hi @ATheorell ,

I want to thank you for the on-time help. It's evident that all your enhancements, especially around optimizing the file retrieval from the toml file and its encapsulation for reusability, are well-thought-out.

1. **File Retrieval Optimization:** I noticed that changes in `file_selector.py` did not alter the selection logic but optimized the process of retrieving files from the toml and encapsulated it for reusability. Could you confirm if my understanding here is correct?

2. **Necessity of main.py in Tests:** I've observed the addition of `(p / "main.py").write_text("The program will be written in this file")` in the test suite. Could you elaborate on the necessity of explicitly creating a main file within the test environment? Previously, the tests ran without this specific file being mentioned. Is this change linked to a new requirement or testing strategy?

Looking forward to your insights on these points.

gpt_engineer/applications/cli/file_selector.py
```diff
@@ -336,112 +275,143 @@ def ask_for_files(project_path: Union[str, Path]) -> FilesDict:
...
+
+    return get_files_from_toml(input_path, toml_file)
+
+
+def get_files_from_toml(input_path, toml_file):
+    selected_files = []
+    edited_tree = toml.load(toml_file)
+    for file, properties in edited_tree["files"].items():
+        if properties.get("selected", False):
+            selected_files.append(file)
```
ATheorell, line 398: Here I changed to the relative path from the project directory

gpt_engineer/applications/cli/file_selector.py
```diff
@@ -336,112 +275,143 @@ def ask_for_files(project_path: Union[str, Path]) -> FilesDict:
...
-            with open(os.path.join(project_path, file.strip()), "r") as content:
-                content_dict[file.strip()] = content.read()
+            selected_files = editor_file_selector(project_path, True)
+    content_dict = {}
+    for file_path in selected_files:
+        # selected files contains paths that are relative to the project path
+        file_path = Path(file_path)
+        try:
+            # to open the file we need the path from the cwd
+            with open(Path(project_path) / file_path, "r") as content:
```
ATheorell, line 304: Since file path is now relative to the project directory, we have to add the project path to read the right file

ATheorell:
Thanks for the praise @similato87 ! 

1. I made a meaningful change inside the file selector (see my inline comments). Using the path relative to the project path is more in line with the project philosophy, but, most of all, it solved the caching issue. Previously, an absolute path was returned from the file selector, which propagated to the prompt. In the tests, we have tmp paths, so the tmp path that was given in the prompt changed every run, which caused problems with caching.

2. I changed the test slightly, by explicitly writing a toml file in the test. This way, I thought the test is slightly closer to the actual use case. This toml file needed to contain something, which is the reason why I invented the main file. The main file is not strictly necessary. If you have a better way of making this test feel free to do it.

As, I said, if you want we can continue in #937 , and you don't need to integrate everything in my commit if you don't like it.

Best,
Axel

similato87:
> Thanks for the praise @similato87 !
> 
> 1. I made a meaningful change inside the file selector (see my inline comments). Using the path relative to the project path is more in line with the project philosophy, but, most of all, it solved the caching issue. Previously, an absolute path was returned from the file selector, which propagated to the prompt. In the tests, we have tmp paths, so the tmp path that was given in the prompt changed every run, which caused problems with caching.
> 2. I changed the test slightly, by explicitly writing a toml file in the test. This way, I thought the test is slightly closer to the actual use case. This toml file needed to contain something, which is the reason why I invented the main file. The main file is not strictly necessary. If you have a better way of making this test feel free to do it.
> 
> As, I said, if you want we can continue in #937 , and you don't need to integrate everything in my commit if you don't like it.
> 
> Best, Axel

Hi Axel,

Your recent commit significantly optimizes the file selection process and encapsulates the functionality for reusability. Please feel free to merge the current PR. Your changes are beneficial and I believe they address potential bugs effectively. Post-merge, I'll focus on enhancing the documentation and annotations for `file_selector` and `test_main`.

To confirm my understanding and for future reference, the bug was primarily due to the use of absolute paths in the file selector, which conflicted with the changing temporary paths in tests, leading to caching and path resolution issues. Adopting a consistent approach using relative paths, as you've implemented, not only resolves this issue in testing but also streamlines the path handling across different environments.

Best regards,
Talion

ATheorell:
Yes, this is the idea with the change!

So, instead of me merging this, I think its fair that you take my changes and continue on #937, so that in the end, your and not my PR gets merged. I think you deserve having a PR merged for all your contributions @similato87 

I hereby close this PR

similato87:
> Yes, this is the idea with the change!
> 
> So, instead of me merging this, I think its fair that you take my changes and continue on #937, so that in the end, your and not my PR gets merged. I think you deserve having a PR merged for all your contributions @similato87
> 
> I hereby close this PR

Hi Axel,

Thank you for your generous acknowledgment and guidance throughout this process and I deeply appreciate the opportunity to integrate these changes into my ongoing work.

I will review all annotations and enhancements thoroughly and continue the work in PR #937 as suggested. Looking forward to contributing more to this collaborative effort!

Best,
Talion

PR #952, opened 2024-01-03
Update chat_to_files.py

Wheaties466:
I ran into issues where i was moving the file around and renaming it but renamed it back before i launched gpt engineer and ran into it an error because the file was not in the file_dict. This will address that. I am also making a comment in this issue https://github.com/gpt-engineer-org/gpt-engineer/issues/928

### DIFF
--- a/gpt_engineer/core/chat_to_files.py
+++ b/gpt_engineer/core/chat_to_files.py
@@ -94,6 +94,13 @@ def overwrite_code_with_edits(chat: str, files_dict: FilesDict):
         The code object to apply edits to.
     """
     edits = parse_edits(chat)
+    for edit in edits:
+        filename = edit.filename
+        if filename not in files_dict:
+            # Log a verbose error message and return gracefully rather than failing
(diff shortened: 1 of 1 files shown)


Wheaties466:
I dont believe this should be failing. its possible the CI/CD tool isnt passing the filename.

Wheaties466:
I know this is closed but i am running into this quite a bit and it tends to eat up some $$ every time this happens. Is there anything i can do to prevent these from happening @ATheorell ?


`WARNING:gpt_engineer.core.chat_to_files:While applying an edit to `New-VMClone2.ps1`, the code block to be replaced was not found. No instances will be replaced`

ATheorell:
Good question: 
1) All code that you paid for is saved in the .gpteng/memory subdirectory of you project. If the paste fails, you can paste in the revised code manually
2) To improve the situation, you can implement PR https://github.com/gpt-engineer-org/gpt-engineer/issues/869

PR #959, opened 2024-01-05
Support custom openai base url

reason211:
OPENAI_BASE_URL=http://127.0.0.1:8080/v1 OPENAI_API_KEY=test python xxxxxx.py


ATheorell:
I never delved into this part of the code really. @pbharrin @UmerHA any reason not to merge this?

ErikBjare:
I thought this was done by the openai package itself.

I searched and found this: https://github.com/openai/openai-python/issues/745

The env var support was removed in some version, and later reintroduced. This should not be needed on latest openai version, it should just work.

Issue #841, opened 2023-11-07
ValueError: Could not parse following text as code edit:


RiccardoRomagnoli:
## Expected Behavior

Improve the code

## Current Behavior

Error gets thrown

## Failure Information

Traceback (most recent call last):

  File "/home/riccardo/.local/bin/gpt-engineer", line 8, in <module>
    sys.exit(app())

  File "/home/riccardo/.local/lib/python3.10/site-packages/gpt_engineer/cli/main.py", line 169, in main
    messages = step(ai, dbs)

  File "/home/riccardo/.local/lib/python3.10/site-packages/gpt_engineer/core/steps.py", line 588, in improve_existing_code
    overwrite_files_with_edits(messages[-1].content.strip(), dbs)

  File "/home/riccardo/.local/lib/python3.10/site-packages/gpt_engineer/core/chat_to_files.py", line 219, in overwrite_files_with_edits
    edits = parse_edits(chat)

  File "/home/riccardo/.local/lib/python3.10/site-packages/gpt_engineer/core/chat_to_files.py", line 268, in parse_edits
    return parse_all_edits(llm_response)

  File "/home/riccardo/.local/lib/python3.10/site-packages/gpt_engineer/core/chat_to_files.py", line 255, in parse_all_edits
    edits.append(parse_one_edit(current_edit))

  File "/home/riccardo/.local/lib/python3.10/site-packages/gpt_engineer/core/chat_to_files.py", line 240, in parse_one_edit
    raise ValueError(f"Could not parse following text as code edit: \n{text}")

### Steps to Reproduce

I'm using this prompt:

> Improve Code for Readability and Reusability:

> 
> Refactor complex functions into smaller, more manageable pieces.
> Use meaningful variable and function names that clearly indicate their purpose.
> Follow a consistent coding style and adhere to best practices outlined in the project's style guide.
> Implement design patterns where applicable to promote code reusability.
> Implement TODOs Where Appropriate:
> 
> Review the codebase for any // TODO: comments and prioritize their completion based on the project's goals.
> Assess the impact of each TODO on the current codebase and potential future developments.
> Document the reasoning behind the resolution of TODOs for future reference.
> Add Comments Where Appropriate:
> 
> Provide clear and concise comments for complex code blocks to explain the logic and its purpose.
> Update or remove outdated comments that no longer reflect the current state of the code.
> Use comments to outline the steps of complex algorithms or workflows within the code.
> 
> Optimize Performance:
> 
[...]

PR #965, opened 2024-01-14
Diff syntax for improve command

similato87:
# Update to File Handling and Improve Preprompt Parsing

## Overview
This PR introduces significant updates to the `file_to_chat` and `file_dict` modules in the `gpt-engineer` project, along with aligning the improve function with the new preprompt guidelines.

### file_dict.py
- Updated the `FilesDict` class's `to_chat` method to format file contents in the new diff syntax as required by the latest improve preprompt.
- Ensured that each file's content is outputted in a clear and unambiguous manner, adhering to the diff syntax with line numbers and +/- indicators.

### chat_to_files.py
- Refined the `parse_edits` function to accurately parse and handle the new diff format from the chat content.
- The function now effectively segregates the 'before' and 'after' parts of each edit and associates them with the corresponding filename.
- Additional error handling and logging are added to ensure robustness in edit parsing.

### Improve Preprompt Update
- The preprompt for the `--improve` function has been revised to conform with the new diff syntax requirements.
- Planning and Output sections are updated to ensure clear and precise guidance for code changes.
- Examples provided in the preprompt illustrate the correct use of the new diff format for adding, modifying, and deleting code.

## Motivation
- These changes align with Issue#869: https://github.com/gpt-engineer-org/gpt-engineer/issues/869 , we aim to enhance clarity and reduce errors in code improvement workflows.
- Also fix other issue caused by this, such as Issue#841: https://github.com/gpt-engineer-org/gpt-engineer/issues/841


similato87:
Hi @ATheorell 

The updates to `file_dict` and the improve preprompt seem to be functioning well. However, I'm encountering issues with parsing the chat content back into the file structure. 

When testing the improved function on a problematic `calculator.py`, the output seems appropriate in terms of planning and adjustments. 
```
PLANNING:
The provided calculator.py file contains a Calculator class with methods that are incorrectly implemented. The add method is performing multiplication, the subtract method is performing division, the multiply method is using a bitwise AND operation, and the divide method has an incorrect conditional check for division by zero. We need to correct these methods to perform their intended operations. Here's the output from the console for reference:
Additionally, the error message in the divide method is incorrect. It should state "Cannot divide by zero" instead of "Cannot divide by two."
OUTPUT:
File: calculator.py
2 -     def add(self, a, b):
3 -         return a * b
2 +     def add(self, a, b):
3 +         # Returns the sum of a and b
3 +         return a + b
5 -     def subtract(self, a, b):
6 -         return a / b
5 +     def subtract(self, a, b):
6 +         # Returns the difference of a and b
6 +         return a - b
8 -     def multiply(self, a, b):
9 -         return a & b
8 +     def multiply(self, a, b):
9 +         # Returns the product of a and b
9 +         return a * b
11 -     def divide(self, a, b):
12 -         if b == 2:
13 -             raise ValueError("Cannot divide by zero.")
11 +     def divide(self, a, b):
12 +         # Raises an error if b is zero and returns the quotient of a and b otherwise
12 +         if b == 0:
13 +             raise ValueError("Cannot divide by zero")
14 +         return a / b
```
However, the conversion of these planned edits back into the code file isn't working as expected. I suspect the issue may be related to the recent changes in this commit: [b9e0e86](https://github.com/gpt-engineer-org/gpt-engineer/pull/965/commits/b9e0e86c5bd63831649e343d215210d8bc3df983). I plan to dive deeper into this tomorrow to identify and fix the root cause.

similato87:
Hi @ATheorell,

I've made significant progress with the code parsing functionality. Using the `parse_edits` function, I've successfully parsed the chat content into structured edits. Here's a quick summary of what I extracted:
```
Skipped line: OUTPUT:
Before edit: ['def add(self, a, b):', 'return a * b']
After edit: ['def add(self, a, b):', '# Perform addition', 'return a + b']
Before edit: ['def subtract(self, a, b):', 'return a / b']
After edit: ['def subtract(self, a, b):', '# Perform subtraction', 'return a - b']
Before edit: ['def multiply(self, a, b):', 'return a & b']
After edit: ['def multiply(self, a, b):', '# Perform multiplication', 'return a * b']
Before edit: ['def divide(self, a, b):', 'if b == 2:', 'raise ValueError("Cannot divide by zero.")']
After edit: ['def divide(self, a, b):', '# Perform division with error handling for division by zero', 'if b == 0:', 'raise ValueError("Cannot divide by zero.")', 'return a / b']
```

I'm currently working on the `apply_edits` function to integrate these edits into the existing code. This should be the last primary step to make it run. It's running successfully in several instances but not consistently. 

One notable observation is that the chat-generated content from the LLM sometimes includes changes to a "prompt" file, and even attempts to delete it. While I've managed to filter these out during parsing, I think a more efficient approach might be to instruct the LLM not to apply changes to the "prompt" file in its pre-prompt phase. This could streamline the process and prevent unnecessary edits from being generated.

ATheorell:
Interesting!
I wonder where it get's the idea of changing the prompt. One thing that may be helpful is to print the full stack of messages that is sent to the LLM when it suggests changing the prompt file. This should make it easier for us to guess what it is in the prompt that has this strange side effect.

ATheorell:
A thing I thought about regarding the parsing. The way to remove lines when parsing is probably this:

1) look at a group of suggested line numbers with a '-' that the LLM suggests.
2) go to the corresponding lines in the code.
3) check if the content in the code is roughly the same as the *before* code in the diff from the LLM, if not, throw warning that edit is infeasible. Otherwise, remove all lines with minus and add all lines with +.

The advantage with this, over exchanging lines based directly on the "before" code that the LLM produces, is that small inconsistencies in the before code may lead to a mismatch. To the contrary, the line number is an unambiguous guide to what line to replace.

similato87:
Hi Axel,

I've just successfully used new improve preprompt to transform a `Calculator` class with incorrect implementations into a fully functional version. 

Here's a quick overview of the transformation:

**Original Code:**
```python
class Calculator:
    def add(self, a, b):
        return a * b

    def subtract(self, a, b):
        return a / b

    def multiply(self, a, b):
        return a & b

    def divide(self, a, b):
        if b == 2:
            raise ValueError("Cannot divide by zero.")
        return a / b
```
Transformed Code:
```python
class Calculator:
    def add(self, a, b):
        return a + b  # Corrected to perform addition

    def subtract(self, a, b):
        return a - b  # Corrected to perform subtraction

    def multiply(self, a, b):
        return a * b  # Corrected to perform multiplication

    def divide(self, a, b):
        if b == 0:  # Corrected to check for division by zero
            raise ValueError("Cannot divide by zero.")
        return a / b
```
This marks the first successful run of our new approach. I plan to delve into resolving the issue with deleting prompts and will also set up more test conditions tomorrow to further validate and refine our implementation.

ATheorell:
Its cool to see the day by day progress! Great that you chose an example with multiple diffs.
Looking forward to merging this :)

similato87:
Hi @ATheorell ,

I'm happy to share that our `improve` function has made advancements. It can now simultaneously handle a variety of complex tasks:

1. **Strict Adherence to Prompts:** The function can interpret and execute instructions with precision, ensuring that the output closely aligns with the given prompt.

2. **Applicability to Both New and Existing Files:** It's now equipped to apply changes to both newly created files and existing ones, enhancing its versatility.

3. **Automatic Error Correction:** The function can autonomously identify and rectify common errors, both syntactical and logical, without the need for specific prompts.

However, there are still some potential inefficiencies. For instance, issues like incorrect indentation(Only if the input file has this issue) or unintended attempts to delete the prompt file may occur, though the probability of these issues is relatively low, approximately 1 in 5.

**Request for Testing and Feedback:**
Axel, would you be able to test this commit when you have time? I would appreciate your feedback. :)

# My lastest run:
## Input 2 files:
**`prompt`:**
```
A very simple calculator. I need two python files, one named calculator for function and the other named main for interaction.
```
**`calculator.py`:** 
```python
class Calculator:
    def add(self, a, b):
        return a - b  # Logical

    def subtract(self, a, b)
        return a - b  # Syntax

    def multiply(self, a, b):
        return lambda a, b: sum(a for _ in range(abs(b))) * (1 if b >= 0 else -1) #unncessary

    def divide(self, a, b):
        if b == 0:
            raise ValueError("Cannot divide by zero.")
        return a / b

    def modulo(self, a, b):
        if b = 0:  # Syntax
            return None  # Logical
        return a % b

    def power(self, a, b):
        result = 1
        for i in range(b):  # Logical
            result *= a
        return result
```
## Output 3 files:
**Improved `calculator.py`:**
```python
class Calculator:
    def add(self, a, b):
        return a + b  # Fixed logical error
    def subtract(self, a, b):  # Fixed syntax error
        return a - b
    def multiply(self, a, b):
        return a * b  # Simplified multiplication
    def divide(self, a, b):
        if b == 0:
            raise ValueError("Cannot divide by zero.")
        return a / b
    def modulo(self, a, b):
        if b == 0:  # Fixed syntax error
            raise ValueError("Cannot modulo by zero.")  # Fixed logical error
        return a % b
    def power(self, a, b):
        result = 1
        for i in range(abs(b)):  # Fixed logical error for negative exponents
            result *= a
        return result
```

**New File `main.py`:**
```python
from calculator import Calculator

def main():
    calc = Calculator()
    while True:
        print("Operations: add, subtract, multiply, divide, modulo, power, exit")
        operation = input("Choose an operation: ").strip().lower()
        if operation == 'exit':
            break
        a = float(input("Enter the first number: "))
        b = float(input("Enter the second number: "))
        try:
            result = getattr(calc, operation)(a, b)
            print(f"The result is: {result}")
        except (ValueError, AttributeError) as e:
            print(e)

if __name__ == "__main__":
    main()
```

### Part of consoles output:
```shell
PLANNING:
To fulfill the request, we need to make several changes to the `calculator.py` file to correct logical and syntax errors. Additionally, we will create a new file named `main.py` for user interaction with the calculator functions.

1. In the `add` method, change the subtraction operator to the addition operator to correct the logical error.
2. Fix the syntax error in the `subtract` method by adding a colon at the end of the method definition.
3. Replace the lambda function in the `multiply` method with a simple multiplication operation, which is the correct and efficient way to perform multiplication.
4. In the `modulo` method, change the single equals sign to a double equals sign to correct the syntax error in the conditional statement.
5. Change the return value from `None` to raising a `ValueError` in the `modulo` method to maintain consistency with the `divide` method error handling.
6. In the `power` method, correct the logical error by checking if `b` is negative and then using the `abs` function to handle the negative exponent correctly.

After correcting the `calculator.py` file, we will create a new file `main.py` that will import the `Calculator` class and allow the user to interact with it. The `main.py` file will contain a simple command-line interface where users can choose an operation and input two numbers to perform that operation using the `Calculator` class.

OUTPUT:

File: calculator.py
2 -     def add(self, a, b):
3 -         return a - b  # Logical
2 +     def add(self, a, b):
3 +         return a + b  # Fixed logical error

5 -     def subtract(self, a, b)
5 +     def subtract(self, a, b):  # Fixed syntax error

8 -     def multiply(self, a, b):
9 -         return lambda a, b: sum(a for _ in range(abs(b))) * (1 if b >= 0 else -1) #unncessary
8 +     def multiply(self, a, b):
9 +         return a * b  # Simplified multiplication

17 -         if b = 0:  # Syntax
18 -             return None  # Logical
17 +         if b == 0:  # Fixed syntax error
18 +             raise ValueError("Cannot modulo by zero.")  # Fixed logical error

21 -     def power(self, a, b):
23 -         for i in range(b):  # Logical
21 +     def power(self, a, b):
23 +         for i in range(abs(b)):  # Fixed logical error for negative exponents

File: main.py
1 + from calculator import Calculator
2 +
3 + def main():
4 +     calc = Calculator()
5 +     while True:
6 +         print("Operations: add, subtract, multiply, divide, modulo, power, exit")
7 +         operation = input("Choose an operation: ").strip().lower()
8 +         if operation == 'exit':
9 +             break
10 +         a = float(input("Enter the first number: "))
11 +         b = float(input("Enter the second number: "))
12 +         try:
13 +             result = getattr(calc, operation)(a, b)
14 +             print(f"The result is: {result}")
15 +         except (ValueError, AttributeError) as e:
16 +             print(e)
17 +
18 + if __name__ == "__main__":
19 +     main()

Please note that the `main.py` file is a new file and should be created in the same directory as `calculator.py`. The `main` function provides a simple command-line interface for the calculator, and the `if __name__ == "__main__":` block ensures that `main` is called when the script is executed directly.Deleted from calculator.py, line 2: 'def add(self, a, b):'
Added to calculator.py, line 2: 'def add(self, a, b):'
Deleted from calculator.py, line 3: 'return a - b  # Logical'
Added to calculator.py, line 3: 'return a + b  # Fixed logical error'
Deleted from calculator.py, line 5: 'def subtract(self, a, b)'
Added to calculator.py, line 5: 'def subtract(self, a, b):  # Fixed syntax error'
Deleted from calculator.py, line 8: 'def multiply(self, a, b):'
Added to calculator.py, line 8: 'def multiply(self, a, b):'
Deleted from calculator.py, line 9: 'return lambda a, b: sum(a for _ in range(abs(b))) * (1 if b >= 0 else -1) #unncessary'
Added to calculator.py, line 9: 'return a * b  # Simplified multiplication'
Deleted from calculator.py, line 17: 'if b = 0:  # Syntax'
Added to calculator.py, line 17: 'if b == 0:  # Fixed syntax error'
Deleted from calculator.py, line 18: 'return None  # Logical'
Added to calculator.py, line 18: 'raise ValueError("Cannot modulo by zero.")  # Fixed logical error'
Deleted from calculator.py, line 21: 'def power(self, a, b):'
Added to calculator.py, line 21: 'def power(self, a, b):'
Deleted from calculator.py, line 23: 'for i in range(b):  # Logical'
Added to calculator.py, line 23: 'for i in range(abs(b)):  # Fixed logical error for negative exponents'
Created new file: main.py
Added to main.py, line 1: 'from calculator import Calculator'
Added to main.py, line 2: 'def main():'
Added to main.py, line 3: 'calc = Calculator()'
Added to main.py, line 4: 'while True:'
Added to main.py, line 5: 'print("Operations: add, subtract, multiply, divide, modulo, power, exit")'
Added to main.py, line 6: 'operation = input("Choose an operation: ").strip().lower()'
Added to main.py, line 7: 'if operation == 'exit':'
Added to main.py, line 8: 'break'
Added to main.py, line 9: 'a = float(input("Enter the first number: "))'
Added to main.py, line 10: 'b = float(input("Enter the second number: "))'
Added to main.py, line 11: 'try:'
Added to main.py, line 12: 'result = getattr(calc, operation)(a, b)'
Added to main.py, line 13: 'print(f"The result is: {result}")'
Added to main.py, line 14: 'except (ValueError, AttributeError) as e:'
Added to main.py, line 15: 'print(e)'
Added to main.py, line 16: 'if __name__ == "__main__":'
Added to main.py, line 17: 'main()'
```

ATheorell:
Hi, I fixed the tests. The functioning version is here: https://github.com/gpt-engineer-org/gpt-engineer/tree/diff-syntax-for-improve-command
I'm to blame for merging Antons PR without properly checking all consequences (which were the main source of problems we experienced here).

ATheorell:
btw, this function (residing in core/default/steps.py) must be changed to have a relaxed criterion for checking whether an edit is admissible:

def incorrect_edit(files_dict: FilesDict, chat: str) -> List[str,]:
    problems = []
    try:
        edits = parse_edits(chat)
    except ValueError as problem:
        print("Not possible to parse chat to edits")
        problems.append(str(problem))
        return problems

    for edit in edits:
        # only trigger for existing files
        if edit.filename in files_dict:
            if edit.content not in files_dict[edit.filename] and edit.is_before is True:
                problems.append(
                    "This section, assigned to be exchanged for an edit block, does not have an exact match in the code: "
                    + edit.content
                    + "\nThis is often a result of placeholders, such as ... or references to 'existing code' or 'rest of function' etc, which cannot be used the HEAD part of the edit blocks. Also, to get a match, all comments, including long doc strings may have to be reproduced in the patch HEAD"
                )
    return problems
    
 Tee problem is that it will flag an edit as soon as we don't have an exact match (though we no longer require and exact match). We should have a general matching function so we test the same thing both here and in the apply_edits function.

gpt_engineer/core/chat_to_files.py
```diff
@@ -100,97 +100,88 @@ def overwrite_code_with_edits(chat: str, files_dict: FilesDict):
...
+    # Sort edits such that 'before' edits come before 'after' edits
+    edits.sort(key=lambda edit: (edit.filename, edit.line_number, not edit.is_before))
     return edits
 
 
 def apply_edits(edits: List[Edit], files_dict: FilesDict):
-    """
-    Apply a list of edits to the given code.
+    # Helper function to compare strings ignoring spaces
+    def is_similar(str1, str2):
```
ATheorell, line 134: I think this criterion could be relaxed further. For example, you could count the number each  character occurs (unordered) and then require that the intersection between the code in the file and the edit needs to be 90% the lengths of the line in the file or something like that. For short lines (like 1 character), the criterion should require a perfect match. This criterion should be called both by apply edits and in the incorrect_edit function that I referenced in the other comment.

similato87:
Hi Axel,

Thanks for your time! All fixed. I will do more tests and write a detailed explanation and demonstration for other team members tomorrow.

gpt_engineer/core/chat_to_files.py
```diff
@@ -100,97 +100,104 @@ def overwrite_code_with_edits(chat: str, files_dict: FilesDict):
...
-            raise ValueError(f"Could not parse following text as code edit: \n{text}")
-        before, after = splits
+# Helper function to compare strings ignoring spaces
+def is_similar(str1, str2):
+    # Remove spaces and convert to lowercase for case-insensitive comparison
+    str1, str2 = str1.replace(" ", "").lower(), str2.replace(" ", "").lower()
 
-        before = before.replace(HEAD, "").strip()
-        after = after.replace(UPDATE, "").strip()
+    # For single character strings, require exact match
```
ATheorell, line 137: Shouldn't this part be superfluous? I believe the nice code below should imply this for all strings shorter than 10 characters? This should be added as a test case and if its passes, these three lines can be removed.

ATheorell:
Last but not least, you mentioned indentation errors? This goes a little beyond the scope of this PR, but it could maybe be a good time to make a general stab at indentation errors. We could use black to format the code, before printing to file in file_store.py upload. We could use something like:

import black
from black import FileMode

code = """
def my_function():
    return 1
"""

# Format the outer code
formatted_outer_code = black.format_str(code, mode=FileMode())

This should fix all indentation errors and not just for improve mode. The only downside is that we have to move black from being a dev dependecny to an common dependency

ATheorell:
That was a quite some feedback! As usual you've done extraordinary work, which gave me the thoughts outlined above. Don't feel forced to implement my suggestions exactly, maybe they are not the best, but they make sense to me :)

similato87:
> Last but not least, you mentioned indentation errors? This goes a little beyond the scope of this PR, but it could maybe be a good time to make a general stab at indentation errors. We could use black to format the code, before printing to file in file_store.py upload. We could use something like:
> 
> ```
> import black
> from black import FileMode
> 
> code = """
> def my_function():
>     return 1
> """
> 
> # Format the outer code
> formatted_outer_code = black.format_str(code, mode=FileMode())
> ```
> 
> This should fix all indentation errors and not just for improve mode. The only downside is that we have to move black from being a dev dependecny to an common dependency

Hi Axel,

I totally agree that using black for code formatting is a more effective approach. Now the `improve`model relies on GPT to generate formatted code, which can be resource-intensive and may not always correctly handle indentation errors. Additionally, this method increases the complexity of the preprompt.

Given that we're considering moving black from a development dependency to a common dependency, it would be beneficial to fully utilize it for both improved mode and regular code generation. This will ensure consistency in code formatting across all outputs.

To streamline our development process, I propose creating a separate PR specifically for integrating black into our normal code generation process. I can initiate this by writing an issue to outline the feature and its implementation details.

similato87:
There are still some boundary issues encountered during testing, especially with more complex tasks. We're optimistic about resolving these by tomorrow. I've drafted a summary of this PR and plan to tag others for review tomorrow.

similato87:
migrated to PR #1005 with new commits

PR #984, opened 2024-01-21
update openai

TheoMcCabe:


pyproject.toml
```diff
@@ -19,15 +19,16 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.dependencies]
 python = ">=3.10,<3.12" # updated for llama-index dependency
-openai = "0.28"
+openai = ">=1.0.0"
```
ErikBjare, line 22: Should probably be ^1.0

pyproject.toml
```diff
@@ -19,15 +19,16 @@ build-backend = "poetry.core.masonry.api"
...
 termcolor = "2.3.0"
 typer = ">=0.3.2"
 rudder-sdk-python = ">=2.0.2"
 dataclasses-json = "0.5.7"
 tiktoken = ">=0.0.4"
 tabulate = "0.9.0"
 python-dotenv = ">=0.21.0"
-langchain = ">=0.0.335"
+langchain = ">=0.1.0"
```
ErikBjare, line 30: Should also probably be `^0.1` (depending on langchain versioning strategy)

ErikBjare (commented):
Noticed we use a lot of `>=` version specifiers that should probably be `^`.

This is never a problem for devs (who use dependency locks) but will be a problem for end-users who install via pip, as that might get them a newer version than supported.

While semantic versioning isn't as accepted in Python-land as in JS-land, using `^` has in my experience helped avoid such issues.

pyproject.toml
TheoMcCabe: Nice one thanks. Will update

TheoMcCabe: done

gpt_engineer/core/ai.py
```diff
@@ -18,7 +18,8 @@
     messages_from_dict,
     messages_to_dict,
 )
-from langchain_community.chat_models import AzureChatOpenAI, ChatOpenAI
+from langchain_community.chat_models import AzureChatOpenAI
```
ErikBjare, line 21: When looking up [docs for langchain_community.chat_models.AzureChatOpenAI](https://api.python.langchain.com/en/latest/chat_models/langchain_community.chat_models.azure_openai.AzureChatOpenAI.html#) they state that it is deprecated.

The [langchain-openai README](https://pypi.org/project/langchain-openai/) state that you should use:

```
from langchain_openai import AzureOpenAI
```

pyproject.toml
```diff
@@ -19,15 +19,16 @@ build-backend = "poetry.core.masonry.api"
...
 termcolor = "2.3.0"
 typer = ">=0.3.2"
 rudder-sdk-python = ">=2.0.2"
 dataclasses-json = "0.5.7"
 tiktoken = ">=0.0.4"
 tabulate = "0.9.0"
 python-dotenv = ">=0.21.0"
-langchain = ">=0.0.335"
+langchain = "^0.1"
+langchain_openai = "^0.0.2"
```
ErikBjare, line 31: My bad before. Given it's still on a pre-`0.1` release, adding a `^` will basically be the same as `==0.0.2`.

Given this is an extra for langchain, we can probably rely on it being correctly resolved with its openai and langchain version requirements.

```suggestion
langchain_openai = "*"
```

gpt_engineer/core/ai.py
TheoMcCabe: So i just looked into making this change. While it seems simple enough I have no way to test it works as i dont have an azure account with an azure open ai set up or the time to set this up. Its possible the API of AzureOpenAI is different to AzureChatOpenAI, and without actually being able to verify the new code works id rather not make this change as part of this PR - but i can create a seperate ticket for it?

ErikBjare: Oops, I wrote `AzureOpenAI` but meant `AzureChatOpenAI` (edited my comment).

I'm fine with it either way, but I think this is a safe change to make without testing.

I have an Azure account that I could test with, but feel free to wait with that for another PR :)

TheoMcCabe: Ah cool I was looking at azure so . I'll do this then

TheoMcCabe: Done [3e83dfc](https://github.com/gpt-engineer-org/gpt-engineer/pull/984/commits/3e83dfc2cf963cd9f127276acd4b08193a6c19df)

ErikBjare (approved):
LGTM!

MrDaVinC:
p

pyproject.toml
MrDaVinC: o

AntonOsika:
🚀

PR #986, opened 2024-01-23
Fix docs build pipeline

kristiankyvik:
This [commit](https://github.com/gpt-engineer-org/gpt-engineer/commit/e28751f7ce2c943338a460cbf91af19050629348) seems to have broken the build pipeline for the docs. The builds have been failing [for the last 6 weeks](https://readthedocs.org/projects/gpt-engineer/builds/). 

I have tested my changes locally, and I can successfully build the docs. This PR should fix this issue. I also added the docs/core folder to the .gitignore as I noticed it was generated when building the documentation locally. 

I am not entirely about how the docs are being built and deployed - is this done automatically? If so, how?

Thanks!

PR #987, opened 2024-01-23
build(docs): install docs deps with poetry instead of docs/requiremens.txt

ErikBjare:
Continuing with fixes after https://github.com/gpt-engineer-org/gpt-engineer/pull/986


ATheorell:
great!

PR #988, opened 2024-01-23
Update docs quick overview

kristiankyvik:
My attempt here was to update the [Quick overview](https://gpt-engineer.readthedocs.io/en/latest/intro/quick_overview.html) section of the docs, since it seems to be out of date. 

Here are the summary of the changes:

- Attempted to update the core components to reflect new structure
- Removed references to old classes and functions and updated the file paths
- Added in-page anchors when required

I also documented a command to run the docs with live reload, which will make it easier to work with them locally.


docs/intro/quick_overview.md
```diff
@@ -1,55 +1,59 @@
...
-GPT-Engineer is a project that uses GPT-4 to automate the process of software engineering. It includes several Python scripts that interact with the GPT-4 model to generate code, clarify requirements, generate specifications, and more.
+GPT-Engineer is a project that uses LLMs (such as GPT-4) to automate the process of software engineering. It includes several Python scripts that interact with the LLM to generate code, clarify requirements, generate specifications, and more.
 
 <br>
 
 ## Core Components
-### 1. AI Class (`gpt_engineer/ai.py`)
-The AI class is the main interface to the GPT-4 model. It provides methods to start a conversation with the model, continue an existing conversation, and format system and user messages.
+
+(ai_class)=
```
ErikBjare, line 8: Is this some markdown extension I'm not aware of? Never seen before 😮

kristiankyvik: Me neither! Apparently, its Sphinx way of doing cross-references in the MyST Markdown dialect ([see docs](https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#explicit-targets)).

ATheorell:
Maybe outside the scope of what you want to do in this PR  @kristiankyvik, but feel free to cut down A LOT on the static documentation files (the ones not generated from docstrings). Chances are that they are outdated and it doesn't seem like we are managing to keep non-generated doc files up-to-date.

kristiankyvik:
Hi @ATheorell, 
I think what you suggest makes a lot of sense. I am happy to take a look at the automatically generated docs (API Reference) next I think there. I will open an issue now to suggest what we can improve. 

Could it make sense to merge this PR in the meantime so that at least the main overview is up-to-date?

TheoMcCabe:
Looks good! Thank you @kristiankyvik . Yes would be great if you had time to look at the auto generated docs too

Issue #990, opened 2024-01-24
Improve automatically generated docs


kristiankyvik:
Suggested steps: 
- [x] Ensure Sphinx is correctly parsing the docstrings - PR for this one is [here](https://github.com/gpt-engineer-org/gpt-engineer/pull/991)
- [x] Make sure that most important classes are correctly documented and have examples - PR for this one [here](https://github.com/gpt-engineer-org/gpt-engineer/pull/994)
- [x] Review the static documentation files to clean/remove the ones that we cannot keep up to date - PR for this one [here](https://github.com/gpt-engineer-org/gpt-engineer/pull/996)

@ATheorell does this makes sense to you? Happy to open PRs for each

kristiankyvik:
This can be closed as the last of the three PRs has been merged.

I created a [new issue](https://github.com/gpt-engineer-org/gpt-engineer/issues/1050) to address the docs broken build step.

PR #991, opened 2024-01-25
Ensure sphinx correctly parses docstrings

kristiankyvik:
This addresses task 1 of this issue: https://github.com/gpt-engineer-org/gpt-engineer/issues/990

Here are the main changes: 

- The `api_reference.rst` is now up-to-date after running the `create_api_rst.py` locally and committing the generated files. Not sure it makes sense to commit all these files or if we simply want to generate them when we build the project
- Ensure that we run the `create_api_rst.py` script in the pre_build step. I think this makes sense, but please let me know if you thing differently. I took inspiration from how the [langchain project seems to have solved this](https://github.com/langchain-ai/langchain/blob/9dd7cbb447f9794f27c5d27f522731fe02a6ee84/.readthedocs.yaml#L21)  

Update 1 - Did the two following changes: 
- Bumped sphinx to 5.0.0, after removing non-maintained package `sphinx_panels` that required an older version of sphinx
- Mocked two packages that are installed as experimental, since these where causing an error when building the docs, since the packages were expected to be installed. 

Update 2 - Did the following change:
- Ensures the auto-generated docs are not commited to version control


MrDaVinC (approved):

docs/conf.py
```diff
@@ -65,7 +65,6 @@
     "sphinx.ext.napoleon",
     "sphinx.ext.viewcode",
     "sphinx_copybutton",
-    "sphinx_panels",
```
kristiankyvik, line 68: Please note that I removed this package because I wanted to update sphinx to 5.0.0 since:  
- We were seeing this error in the docs build pipeline (see [link](https://readthedocs.org/projects/gpt-engineer/builds/23230027/)). 
- This package, which was not being uses anywhere, is not maintained anymore ([link](https://sphinx-panels.readthedocs.io/en/latest/)) and impeding us to update to this  more recent version of sphinx.

docs/conf.py
```diff
@@ -86,6 +85,7 @@
 autodoc_member_order = "groupwise"
 autoclass_content = "both"
 autodoc_typehints_format = "short"
+autodoc_mock_imports = ["tree_sitter_languages", "llama-index"]
```
kristiankyvik, line 88: After updating sphinx to 5.0.0, when building the docs, I was running into a different error:

```
Handler <function process_automodsumm_generation at 0x7fedb8d30430> for event 'builder-inited' threw an exception (exception: No module named 'gpt_engineer.tools.experimental.document_chunker')
make: *** [html] Error 2
```

I followed the advice [here](https://stackoverflow.com/questions/68398777/what-is-the-cause-of-this-sphinx-autodoc-mockfinder-error) and ensure that the two files were mocked when building the docs. This fixed the issue.

TheoMcCabe (approved):
looks good thank you

PR #994, opened 2024-01-31
Update / generate the docstrings

kristiankyvik:
This PR generated the docstrings in the NumPy style for all the code. In some cases it also updates outdated docstrings.

As suggested by @ATheorell, I used gpt-engineer CLI to generate all of docstrings. 

This will properly populate the API reference section of the `gpte-engineer` docs (see image below):
 
CleanShot 2024-01-31 at 17 07 14@2x

I think we should try to merge [this PR before](https://github.com/gpt-engineer-org/gpt-engineer/pull/991), that fixes the documentation build before merging this one as described in the [original issue](https://github.com/gpt-engineer-org/gpt-engineer/issues/990).


gpt_engineer/applications/cli/file_selector.py
```diff
@@ -221,10 +316,22 @@ def get_files_from_toml(self, input_path, toml_file):
         return selected_files
 
     def merge_file_lists(
-        self, existing_files: list[str], new_files: list[str]
```
ATheorell, line 224: Out of curiosity, did you do this @kristiankyvik , or was this an automatic type correction?

ATheorell:
Wow, massive PR! Great work. Its so comprehensive I didn't read it all. Is there any reason why we should not merge it @kristiankyvik , if not, I'll click the button :)

gpt_engineer/applications/cli/file_selector.py
kristiankyvik: This was GPT-engineer. Despite me explicitly requesting it to limit itself to adding/editing docstrings, in some cases, it took the liberty of also correcting the parameter types. I figured I would also keep these suggestions in the PR after manually reviewing them.

kristiankyvik:
I see no reason for us not to merge this other than us perhaps merging [this PR first](https://github.com/gpt-engineer-org/gpt-engineer/pull/999) as this should fix the docs build. I already made some changes to the PR based on your feedback from yesterday.

kristiankyvik:
I think this can be merged now, @ATheorell since #999 has been merged. 

We are now down to only 18 errors, all related to static files that will be addresses in the last PR of this series, #996.

PR #999, opened 2024-02-06
Avoid docs build error by removing experimental packages & addressed a lot of the errors and warnings in documentation build

kristiankyvik:
Hi, 

The sphinx documentation [build was failing](https://readthedocs.org/projects/gpt-engineer/builds/23352466/) because of a `exception: no module named xxxx` error caused by an external dependencies (in this case lama-index) not being available at build time and breaking the building process when running: 

```
python -m sphinx -T -b html -d _build/doctrees -D language=en . $READTHEDOCS_OUTPUT/html
```

~This PR fixes this error, by properly mocking the llama-index package.~

**UPDATE:**
I ended up taking @ATheorell suggestion and have tried deleting all the experimental dependencies + code that depends on them.


ATheorell:
Great that you found a solution to this, however, there is a more radical solution possible that I would prefer. Delete the experimental dependencies and all the code that depends on them. We keep having issues like this for the experimental dependencies and I don't even think anybody is using them. You think you could do this?

kristiankyvik:
Yes, I think this makes sense, let me have a look at that.

kristiankyvik:
I added to the scope of this PR to address a lot of the warning messages that were generated in the build. 

We have gone from 164 warnings to 35 warnings. The remaining warnings are related to formatting discrepancies of the docstrings which I expect to be addressed in [PR #994](https://github.com/gpt-engineer-org/gpt-engineer/pulls).

Please find attached the files containing the logs before and after the last changes.

[logs_before.txt](https://github.com/gpt-engineer-org/gpt-engineer/files/14220373/logs_before.txt)
[logs_after.txt](https://github.com/gpt-engineer-org/gpt-engineer/files/14220375/logs_after.txt)

docs/conf.py
```diff
@@ -107,7 +106,7 @@
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
```
kristiankyvik, line 109: Setting language to english as this was raising an error.

docs/conf.py
```diff
@@ -138,7 +137,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ["_static"]
+# html_static_path = ["_static"]
```
kristiankyvik, line 140: Commented out this setting as this was not being used and it was also generating a warning.

pyproject.toml
```diff
@@ -56,14 +56,6 @@ sphinx-copybutton = ">=0.5.1"
 markdown-include = ">=0.6.0"
 sphinx_copybutton = ">=0.5.2"
 
-[tool.poetry.group.experimental]
```
kristiankyvik, line 59: Removed all experimental dependencies + files depending on them

docs/conf.py
```diff
@@ -202,3 +201,5 @@
 myst_enable_extensions = [
     "colon_fence",
 ]
+
+myst_all_links_external = True
```
kristiankyvik, line 205: This ensures the myst_parser parses all links as hyperlinks. This will remove all the warnings related to links pointing to external files.

PR #1015, opened 2024-02-13
Ollama and litellm integration

definitiontv:
This is just an additional docker compose file with support files  and some notes in the Docker readme.
This introduces an optional compose file which uses Ollama and litellm to handle all model management
The ollama section is running the official ollama/ollama docker image - note it seems to be 4 months old as of this post, so we need to keep an eye on it.

The litellm docker image is currently also out of date as of this image and also does not have ARM support
For this reason I have created a seperate docker file which creates a container form the published python binaries

The gpt-engineer container is built in the same way as the original 

review the .env files and the dc.yml compose file for the environmnt variables that are required
Especially note that you can uncomment the ports sections to make these services available on your local machine or remotely. Be careful about security though. Hackers seem to be stealing free inference endppoints.

A few Comments have been left in to show alternate formulations, you can secure litellm using keys see their documentation


viborc:
We are closing this PR since this is adding an extra overhead with Docker. Also, this solution relies on ollama and lite-llm and we decided to avoid it

I am providing this link as an additional context, and @zigabrencic's discussion about lite-llm https://github.com/BerriAI/litellm/issues/2001

PR #1017, opened 2024-02-15
azure openai fix and vim in dockerfile for improve

rolandknight:


rolandknight (commented):
- ai.py fix for azure openai endpoint
- added vim to dockerfile to support improve file edit

definitiontv:
Not a big point but locally I am using nano  - far far easier than vim  - just as compatible in my view.

ATheorell:
I will close this, as the same azure fix was posted in a PR by @Emassei before this PR. Including vim in the docker environment seems like a good idea (no drawbacks), but on its own, its a too small PR. Could potentially be included in #1016

PR #1022, opened 2024-02-18
Improve mode context files should be system message

ATheorell:
In the improve workflow, we add the context files as HumanMessage. I believe that is wrong. Opinions welcome (and benchmarks needed!)


ErikBjare:
I think it's correct as is (this is what humans do when manually copy-pasting into ChatGPT).

Although, arguably, the input files should be in the same HumanMessage as the request (not have two HumanMessage in sequence).

ATheorell:
Fair. I guess what we should do is to benchmark which version is better once the benchmarks are running :)

PR #1024, opened 2024-02-19
Fix CLI help message

hmasdev:
I found that wrong help messages in `gpte --help`.
This pull-request fixes the help messages.


akaihola:
@ATheorell, could you still accept the PR without making @hmasdev a contributor? The help texts for `--clarify` and `--self-heal` are clearly incorrect and confuse users.

PR #1030, opened 2024-02-25
add openai_api_base of env for other openai_api url

lasoqw:
make other openai_api url be able to work by adding openai_api_base=os.getenv("openai_api_base") , like

set openai_api_base=[your api base url] on cmd

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,6 @@ webapp/.next/
 
 #ignore tox files
 .tox
+
+#ignore log files
+*.log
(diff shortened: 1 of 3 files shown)


ATheorell:
Using the environment variable OPENAI_API_BASE is actually already supported by default via langchain. You use this syntax: https://gpt-engineer.readthedocs.io/en/latest/open_models.html# (I wouldn't trust this documentation entry entirely though, we are currently reworking this stuff).

Regarding logging and verbosity: We would love to have a whole overhaul of logging, using a verbose flag which takes all print statements into account somehow. Would you be interested in picking this up @lasoqw ?

PR #1035, opened 2024-02-27
Update README.md

harmeetsingh-work:
Arg -i missing under improve code section.


viborc (commented):
Hey @harmeetsingh-work I think you've correctly identified the issue, but we are not accepting cosmetic PRs (to keep the bar for contributors relatively high)

Because of that, this will be rejected. I'd appreciate it if you opened this as an issue. We will be collecting those, and then other contributors will be able to close them.

Issue #1023, opened 2024-02-19
Feature request: Integration of Code Formatting and Linting Tools for the improve Function Workflow


similato87:
## Policy and Info

- Maintainers will close issues that have been stale for 14 days if they contain relevant answers.
- Adding the label "sweep" will automatically turn the issue into a coded pull request. This works best for mechanical tasks. More info/syntax at: https://docs.sweep.dev/
- Consider adding the label "good first issue" for interesting, but easily approachable features.

## Feature Description

I propose the integration of code formatting and linting tools into the workflow of our `improve` function. This would involve two key stages: pre-processing and post-processing. In the pre-processing stage, before the code is sent to LLMs for improvement, the code would be formatted and linted to ensure it adheres to standard coding practices and is free of syntax errors. In the post-processing stage, after the LLMs have applied changes, the code would again be subjected to formatting and linting to refine the improvements, rectify any introduced errors, and standardize the code style.

## Motivation/Application

The integration of these tools serves several purposes:

1. **Pre-processing checks**: By checking the code for errors before sending it to LLMs, we can reduce the computational load on LLMs, allowing them to focus more on the complexity of the tasks rather than correcting basic syntax errors. This not only enhances efficiency but also improves the quality of the improvements suggested by LLMs.

2. **Post-processing refinement**: After LLMs have applied their changes, a second round of formatting and linting can significantly enhance user experience by ensuring that the improvements adhere to standard coding practices, and any errors inadvertently introduced by LLMs are corrected. This step is crucial for maintaining code quality and readability, especially for dynamically typed languages where indentation and formatting are vital, like Python.

### Suggested Tools:

- **Python**: Utilize **Black** for code formatting and **Pylint** or **Flake8** for linting to ensure adherence to PEP 8 standards and to catch common errors.
- **JavaScript**: Employ **Prettier** for formatting and **ESLint** for linting, allowing for customizable rules to fit our project's coding standards.
- **Java**: Implement **Google Java Format** for formatting to ensure the code is consistent with Google's Java Style Guide.

These tools are widely recognized for their robustness and flexibility, making them suitable for integration into our existing workflow. This feature would not only streamline the development process but also significantly enhance the quality and reliability of code improvements generated through the `improve` function.

Styren:
Hey! Sounds like a useful feature. I had some questions;

[...]

PR #1043, opened 2024-03-04
feat: added `gpt-engineer.toml` project configuration file

ErikBjare:
Here is an implementation of the proposed `gpt-engineer.toml` configuration file as I mentioned in the last meeting and as discussed in https://github.com/gpt-engineer-org/gpt-engineer/issues/1023

Work is needed to fully integrate/implement use of the configuration.

Comments/review welcome!

### DIFF
--- a/gpt_engineer/core/project_config.py
+++ b/gpt_engineer/core/project_config.py
@@ -4,7 +4,7 @@ Functions for reading and writing the `gpt-engineer.toml` configuration file.
 The `gpt-engineer.toml` file is a TOML file that contains project-specific configuration used by the GPT Engineer CLI and gptengineer.app.
 """
 
-from dataclasses import asdict, dataclass
+from dataclasses import asdict, dataclass, field
 from pathlib import Path
 
 import tomlkit
@@ -74,8 +74,8 @@ def filter_none(d: dict) -> dict:
 class Config:
     """Configuration for the GPT Engineer CLI and gptengineer.app via `gpt-engineer.toml`."""
 
-    project: _ProjectConfig = _ProjectConfig()
-    run: _RunConfig = _RunConfig()
+    project: _ProjectConfig = field(default_factory=_ProjectConfig)
+    run: _RunConfig = field(default_factory=_RunConfig)
     gptengineer_app: _GptEngineerAppConfig | None = None
 
     @classmethod
--- a/gpt_engineer/core/project_config.py
+++ b/gpt_engineer/core/project_config.py
@@ -1,0 +1,159 @@
+"""
+Functions for reading and writing the `gpt-engineer.toml` configuration file.
+
+The `gpt-engineer.toml` file is a TOML file that contains project-specific configuration used by the GPT Engineer CLI and gptengineer.app.
+"""
+
+from dataclasses import asdict, dataclass
+from pathlib import Path
+
+import tomlkit
+
+default_config_filename = "gpt-engineer.toml"
+
+example_config = """
+[project]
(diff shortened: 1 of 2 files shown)


ErikBjare:
> Is the idea that the users are writing their own config toml files?

The idea is that powerusers will have a way to specify project-specific overrides, but ordinary/first-time users shouldn't need to know about them.

> Is the idea that they configure in the toml how their generated programs are going to be linted?

Yes. We would have some reasonable defaults, like `ruff` (fast) or `mypy` (for typechecking), but ultimately the user should be able to choose `eslint` for example (if they are using a JS codebase).

We could also make gpt-engineer include such a recommended configuration file for new generations (potentially customized, depending on chosen tech stack/template, as we do in gptengineer.app).

> I see that there is a hurdle to be overcome in documenting and teaching users how to interact with the config.

We have not yet fully integrated the config internally for gptengineer.app, but when we do, I will write documentation for it.

>Is the most important aim here is to make the software more configurable, to make it easier to reuse code between projects (such as cli and app)?

The aim here originally was to make projects able to specify what its toolchain is in a language-agnostic way.

We ended up needing this at `gptengineer.app` because we want to support generic tech stacks and repository structures, and a configuration file like this solves the problem.

We also found a need for the configuration file when we wanted to create a client CLI to our backend, as a helper tool for setting up `gptengineer.app` to work with externally created codebases, and the same would apply to the gpt-engineer CLI.

>In this case, the main aim for the PR on the cli side should be to have great defaults that most users stick to.

:100:

similato87:
Thank you, @ErikBjare, for proposing this solution and sharing the implementation details. I believe it's crucial to establish a standard configuration file at this juncture. With the diverse range of customization preferences our users have, it's impractical to accommodate all possible options through CLI alone. Hence, having configuration files to manage settings for lint, format, and potentially other preferences is essential. I plan to discuss the integration of this configuration into our CLI with Axel, particularly focusing on balancing the functionality of the `improve` function with the need to minimize dependencies in our library packages.

AntonOsika:
Love it.

Any next steps before merging?:)

ATheorell:
Thanks  a lot for the great explanation @ErikBjare. 

Good point with the dependencies @similato87.

For the median CLI user, I imagine that a default that recognizes .py and .js files and lints these with no questions asked would be useful.

I feel like I'm mixing into all PRs currently, which goes beyond my capacity. From my point of view, you can develop and merge this independently of me.

AntonOsika:
Let's merge 🚀

AntonOsika:
I think we should go ahead and use this format for config file at and merge!*

Currently that configurations are done in .gpteng is a pain. It is non-transparent and undocumented.

This is definitely the right direction and standardises things.

gpt_engineer/core/project_config.py
```diff
@@ -0,0 +1,159 @@
...
+from pathlib import Path
+
+import tomlkit
+
+default_config_filename = "gpt-engineer.toml"
+
+example_config = """
+[project]
+base_dir = "./frontend"  # base directory to operate in (for monorepos)
+src_dir = "./src"        # source directory (under the base directory) from which context will be retrieved
```
ErikBjare, line 17: I'm not sure about this section. Would perhaps make more sense with a `paths` section, such that it can be just:

```suggestion
[paths]
base = "./frontend"  # base directory to operate in (for monorepos)
src = "./src"        # source directory (under the base directory) from which context will be retrieved
```

ErikBjare:
Merging this now, happy to talk more about it on the meeting tomorrow, among other exciting topics!

PR #1025, opened 2024-02-21
Benchmark on APPS dataset

azrv:
This PR adds an ability to benchmark gpt-engineer on problems given in [APPS dataset
](https://huggingface.co/datasets/codeparrot/apps)

How to run?
`python3 gpt_engineer/benchmark gpt_engineer/core/default/simple_agent.py apps`

Original Issue: https://github.com/gpt-engineer-org/gpt-engineer/issues/819

PR #1045, opened 2024-03-04
Improved preprompts for improve flow

ATheorell:
I have noticed that the current preprompts for the improve flow sometimes leads to lazy behavior, in the sense that no or little code is written. In this PR, I have aligned the preprompts as much as possible with the generate workflow, which we know is not lazy. I hope this will lead to tangible improvements and hopefully we can soon test the actual improvement using the apps benchmarks which is about to be supported (#1025).


ATheorell:
@ErikBjare is there a quick explanation why the codecov workflow fails?

ErikBjare:
@ATheorell codecov.io has these configurable thresholds for what constitutes an acceptable code coverage.

Since we only use it for informational purposes, and not as a blocker for merging PRs, we don't want the checks to fail.

I was supposed to configure this, but I cannot since I don't have org-level maintainer access.

Here are the steps:

1. Go here: https://app.codecov.io/gh/gpt-engineer-org
2. Go to settings -> Global YAML <details><summary>Screenshot</summary>image</details>
3. Paste this:
```
coverage:
  status:
    patch:
      default:
        informational: true
    project:
      default:
        informational: true
```

ATheorell:
Thanks for instructions. I have now updated the global yaml. Tried to rerun the jobs, but maybe I need to make a new commit for this to come into effect. Anyway, we can still merge if we want.

similato87:
Hi @ATheorell , this looks fantastic! I've noticed instances where the LLMs yield no differences, leading to hit 'no diff' exceptions for users. My workaround has been advising users to reselect files and modify their prompts. However, ensuring that the LLMs consistently offer guidance on changes would enhance user experience, as that's the primary reason users engage with this feature. 
Additionally, introducing an option for users to specify their preference for the extent of changes—via a flag—could align well with the new configuration feature from @ErikBjare. This approach could make the tool more user-centric and adaptable to individual user needs.

gpt_engineer/preprompts/improve
```diff
@@ -1,50 +1,15 @@
...
--
--Every diff must be fenced with ```CONTENT OF EDIT BLOCK``` with the correct code language.
--
--The file names at the beginning of a diff, (lines starting with --- and +++) is the relative path to the file before and after the diff.
--
--LINES TO BE REMOVED (starting with single -) AND LINES TO BE RETAIN (no starting symbol) HAVE TO REPLICATE THE DIFFED HUNK OF THE CODE EXACTLY LINE BY LINE.
--NEVER OMIT ANY WHITESPACE NEEDED TO KEEP THE SUCH LINES IDENTICAL.
--KEEP THE DIFF HUNKS (A HUNK STARTS WITH "@@ LINE NUMBER INSTRUCTIONS @@" AS COMPACT AS POSSIBLE; IF POSSIBLE KEEP HUNKS SHORTER THAN 19 LINES BY BREAKING THEM UP INTO MULTIPLE HUNKS.
--AVOID STARTING A HUNK WITH AN EMPTY LINE
+When you are done, write finish with "this concludes a fully working implementation".
```
TheoMcCabe, line 15: Unsure about this ... is never a fully working implementation because its only outputting a change? is it sometimes a fully working implementation and sometimes not a fully working implementation because its doing the work in chunks ?

ATheorell: I've adopted a flow that's very similar to the generate syntax, to make the diff syntax less "lazy" and in my tests, this prompting is a lot better. The "this concludes a fully working implementation" part is some kind of hack, which makes the LLM more prone to actually write code, rather than a delimiter. Potentially it could be moderated to be more in the "diff" context, but I think it does the job.

Issue #819, opened 2023-10-23
Automatic benchmarking of gpt-engineer with APPS


ATheorell:
## Feature description
gpt-engineer has an automatic evals suite in "evals/eval_new_code.py". However, only 2 test cases are given in evals/new_code_eval.yaml . An alternative to filling in more testcases manually, we should parse in prompts and tests from the (very large) APPS dataset (https://paperswithcode.com/dataset/apps).

Since APPS is way too large to run in its entirety, there should be functionality to run n randomly selected tests and run n tests according to some predetermined test ordering (so that consecutive benchmark runs are comparable). 

The APPS database should not be added to the gpt-engineer git repo! Probably the best way to handle this is to pull it from huggingface (https://huggingface.co/datasets/codeparrot/apps) in the code itself (potentially caching it and gitignoring it so it doesn't need to be pulled on every run).

## Motivation/Application
Automatic benchmarking is the ideal way to determine whether an imposed change to the code base is advantageous.

pbharrin:
I can add a sampled version of the APPS dataset which will give us a good idea how our project is doing without costing a fortune.  APPS is a great at testing how well our repair of broken code works.

AntonOsika:
Added to benchmarking milestone

Would be happy to celebrate getting it done before new years in 2 weeks!

azrv:
@ATheorell assign to me

ATheorell:
Yes, please a shot shot at this @azrv :)

azrv:
https://github.com/gpt-engineer-org/gpt-engineer/pull/1051 was merged 🎉

@pbharrin It's now a matter of cherry-picking problems we want to constantly test against.

[gpt_engineer/benchmark/benchmarks/apps/problems.py:4](https://github.com/gpt-engineer-org/gpt-engineer/blob/main/gpt_engineer/benchmark/benchmarks/apps/problems.py)

PR #1051, opened 2024-03-07
Integrate APPS benchmarking 

azrv:
Integrate APPS dataset for benchmarking.

Changes:
1) I faced a few _git diff_ related errors while parsing LLM's response which I wrapped into `DiffError`. I ignore such error and continue running. I guess that's a temporary change until we polish _git diff_ feature 
2) Also there was `regex` dependency added in order to timeout endless regex matching inside `parse_diffs` function. (done in https://github.com/gpt-engineer-org/gpt-engineer/pull/1067)
3) `run.py` had to be adjusted to work with multiple commands/inputs as APPS might provide multiple inputs and outputs per problem

**How to run?**

`python3 gpt_engineer/benchmark gpt_engineer/core/default/simple_agent.py apps`

I tried to run it on first 50 problems, here is the output (total time is not right, because generated code is cached from previous runs)
```
--- Results ---
Total time: 0.38s
Completely correct tasks: 12/50
Total correct assertions: 229/500
Average success rate: 45.79999999999998% on 50 tasks
--- Results ---
```


gpt_engineer/core/default/self_healing_agent.py
```diff
@@ -0,0 +1,43 @@
...
+
+from gpt_engineer.core.ai import AI
+from gpt_engineer.core.base_memory import BaseMemory
+from gpt_engineer.core.default.steps import improve
+from gpt_engineer.core.files_dict import FilesDict
+from gpt_engineer.core.preprompts_holder import PrepromptsHolder
+
+
+SUCCESS = 0
+NO_RESPONSE = 2
```
AntonOsika, line 18: Can we remove these?

AntonOsika:
Super excited about this, wanted to use it today

I'll add some comments to suggest simplifications.

Apart from that, are we ready to merge or anything you want to fix first? (apart from just resolving pyproject.toml and running `poetry lock`

ATheorell:
I think @azrv wanted to do some cleanup. 

Currently, this PR comes with 2 new agent implementations, the benchmark_agent and the self_healing_agent. For the "base" PR of APPS, probably neither of these should be included; it should be sufficient to run the default_agent, even though performance may be deplorable. It is then the fun of coming PRs to add smart self-healing etc which improves performance.

AntonOsika:
Finally, I don't think need to add inputs:

You can just set the command = None, and then create a dict comprehension of assertions that each loop over all the inputs and calls the command. Makes sense?

ATheorell:
> Finally, I don't think need to add inputs:
> 
> You can just set the command = None, and then create a dict comprehension of assertions that each loop over all the inputs and calls the command. Makes sense?

Ah, the inputs list and assertions list was my invention, but yes, I believe the design where the inputs are stored in the assertions is nicer. I would vouch for that.

gpt_engineer/benchmark/benchmarks/apps/load.py
```diff
@@ -0,0 +1,87 @@
...
+from gpt_engineer.benchmark.benchmarks.apps.problems import PROBLEM_IDS
+from gpt_engineer.benchmark.types import Benchmark, Task, Assertion
+from gpt_engineer.core.files_dict import FilesDict
+from datasets import load_dataset, load_from_disk, Dataset, DatasetDict
+
+DATASET_PATH = "gpt_engineer/benchmark/benchmarks/apps/dataset"
+MAX_N_TEST_EXAMPLES = 10
+
+
+class AppsAssertion:
```
azrv, line 25: Get rid of that

azrv:
> Thanks for PR @azrv
> 
> Some good things here!
> 
> I'll be direct here though: we need higher standards to merge any PR, specifically:
> 
> 1. There is no explanation of why to do things in a new way (unless obvious)
> 2. The PR adds both a new agent to evaluate and a new benchmark. They should be separate PRs.
> 3. The PR completely changes the benchmark code, without explanations of why (and it seemingly adds "agent logic" in the PR?)
> 
> Referring to this in 3: image

Hey!
1) Sure, I updated the PR description
2) Both agents were dropped
3) I cleaned things up and removed 'agent based' logic, that was added solely for self healing in sake of better performance

Here are few things I found could be handy for gpt-engineer benchmarks. Please share your thoughts whether this is smth we would consider doing later: 
1) `retries` AKA self-healing if most of the assertions didn't succeed
2) Running benchmark in multiple threads

azrv:
> Finally, I don't think need to add inputs:
> 
> You can just set the command = None, and then create a dict comprehension of assertions that each loop over all the inputs and calls the command. Makes sense?

Not sure this is achievable with simple dictionaries, because you need to store `command` somewhere.
For now I simply extended `Assertion` class, let me know if it clicks with you or please elaborate on `dict` way

Yes, other benchmarks are still broken. Will fix after we agree on single solution :)

ATheorell:
> > Finally, I don't think need to add inputs:
> > You can just set the command = None, and then create a dict comprehension of assertions that each loop over all the inputs and calls the command. Makes sense?
> 
> Not sure this is achievable with simple dictionaries, because you need to store `command` somewhere. For now I simply extended `Assertion` class, let me know if it clicks with you or please elaborate on `dict` way
> 
> Yes, other benchmarks are still broken. Will fix after we agree on single solution :)

The code is getting more general which is great! I think we can still scale back a lot on the changes to the general run.py file by implementing the system that @AntonOsika suggested. It is not necessary to extend the assertion class. One way (which I think is quite nice), for doing this, is to use the AppsAssertion class that I implemented. The idea goes:

For each input output/pair, create an AppsAssertion object. Pass code, (timeout), command + input, expected output and an execution environment to the constructor of the AppsAssertion object. In the evaluate method, run the code through the execution command and check whether the assertion holds. Timeout and runtime failures will make the assertion fail in the same way as a wrong output (but we could log it differently in the future).

A drawback with this method may be that it complicates parallelization, which could, however, still be added somehow from run.py. Another idea would be to parallelize on problem level. Input on this @AntonOsika ?

Hope this helps

gpt_engineer/benchmark/types.py
```diff
@@ -72,5 +72,14 @@ class Benchmark:
...
     duration: float
+
+    # Returns success rate from 0.00 up to 1.00
+    @property
+    def success_rate(self) -> float:
+        succeeded = len(
+            [result for result in self.assertion_results if list(result.values())[0]]
+        )
+
+        return succeeded / len(self.assertion_results)
```
ATheorell, line 85: I get: "ZeroDivisionError: division by zero" here when I run the first 5 problems.

gpt_engineer/benchmark/run.py
```diff
@@ -48,8 +49,20 @@ def run(
     """
     task_results = []
     for task in benchmark.tasks:
+        print(f"--> Running task: {task.name}\n")
+
         t0 = time.time()
-        files_dict = agent.improve(task.initial_code, task.prompt, task.command)
+        try:
+            files_dict = agent.improve(task.initial_code, task.prompt)
+        except DiffError:  # Temporary catch errors related to git diffs
```
ATheorell, line 57: Is this still a problem? I see that this was necessary during development, but I think we should aim to fix this on the diff side, rather than having a specific catch in the benchmarks.

gpt_engineer/benchmark/benchmarks/apps/load.py
```diff
@@ -0,0 +1,114 @@
...
+        tasks.append(
+            Task(
+                name=str(problem.id),
+                initial_code=FilesDict({"main.py": problem.starter_code}),
+                command=None,  # Explicitly setting `None` because each assertion specifies its command
+                prompt=problem.question
+                + "\nThe program, including its inputs, should be run from the command "
+                "line like 'python main \"input1 input2 etc \"', with all inputs inside "
+                "the quotation marks. The program should not read inputs from stdin.",
+                assertions=[
```
ATheorell, line 94: We can remove the list wrapping again and just use a dictionary comprehension like in the original design

gpt_engineer/benchmark/run.py
```diff
@@ -72,13 +86,17 @@ def run(
         task_results.append(
             TaskResult(
                 task_name=task.name,
-                assertion_results={
-                    assertion_name: assertion(exec_result)
-                    for assertion_name, assertion in task.assertions.items()
-                },
+                assertion_results=[
```
ATheorell, line 89: Same here, we should be able to revert, now that we are using the AppsAssertion design. Each apps assertion will be an entry in the dictionary.

gpt_engineer/benchmark/types.py
```diff
@@ -57,7 +57,7 @@ class Task:
     initial_code: Optional[FilesDict]
     command: Optional[str]
     prompt: str
-    assertions: Optional[Dict[str, Assertion]]
+    assertions: Optional[List[OrderedDict[str, Assertion]]]
```
ATheorell, line 60: Again, revert this

gpt_engineer/benchmark/types.py
```diff
@@ -72,5 +72,14 @@ class Benchmark:
 @dataclass
 class TaskResult:
     task_name: str
-    assertion_results: dict[str, bool]
+    assertion_results: List[dict[str, bool]]
```
ATheorell, line 75: again, revert

gpt_engineer/core/chat_to_files.py
```diff
@@ -134,21 +140,25 @@ def parse_diffs(diff_string: str) -> dict:
     - dict: A dictionary of Diff objects keyed by filename.
     """
     # Regex to match individual diff blocks
-    diff_block_pattern = re.compile(
+    diff_block_pattern = regex.compile(
```
ATheorell, line 143: It is great that you are improving the diff-pipeline! Formally, this should be a separate PR. Would you be able to make a separate PR with the changes to chat_to_files?

ATheorell: I can merge that quickly

pyproject.toml
```diff
@@ -32,6 +32,8 @@ langchain_openai = "*"
 toml = ">=0.10.2"
 pyperclip = "^1.8.2"
 langchain-anthropic = "^0.1.1"
+datasets = "^2.17.1"
+regex = "^2023.12.25"
```
ATheorell, line 36: it would be nice if you can comment what functionality makes the new dependencies necessary

ATheorell (changes requested):
Great work and this is more or less working. Just requesting a bit of housekeeping now really.

gpt_engineer/core/chat_to_files.py
azrv: Will do

azrv: @ATheorell https://github.com/gpt-engineer-org/gpt-engineer/pull/1067

gpt_engineer/benchmark/benchmarks/apps/load.py
```diff
@@ -0,0 +1,114 @@
...
+MAX_N_TEST_EXAMPLES = 10
+
+
+class AppsAssertion:
+    def __init__(self, expected: str, command: str):
+        self.expected_output = self._format(expected)
+        self.command = command
+
+    def evaluate(self, assertable: Assertable) -> bool:
+        pro = assertable.env.popen(self.command)
```
ATheorell, line 35: This may sound a little counterintuitive to the architecture, but I think it may be preferable to create a fresh DiskExecutionEnvironment, upload the code and run from there, instead of using the global one. The reason is that, executing the code may have side-effects like creating and removing files and folders. If this is the case, multiple executions of the same program with different inputs will not be independent.

azrv: I see, will do

gpt_engineer/benchmark/run.py
```diff
@@ -48,8 +48,20 @@ def run(
     """
     task_results = []
     for task in benchmark.tasks:
+        print(f"--> Running task: {task.name}\n")
+
         t0 = time.time()
-        files_dict = agent.improve(task.initial_code, task.prompt, task.command)
+        try:
+            files_dict = agent.improve(task.initial_code, task.prompt)
+        except RecursionError:  # Temporary catch errors related to git diffs
```
azrv, line 56: @ATheorell either diff errors were indeed fixed in latest master or I added them prematurely at the first place, but now the only thing left is `validate_and_correct` function getting into recursion. 

I suggest I temporary catch it here and file an issue for it

azrv:
@ATheorell @AntonOsika All fixed
The only thing left is `RecursionError` explained above

ATheorell (approved):

gpt_engineer/benchmark/run.py
ATheorell: I actually suggest you remove this catch. We merge without it and let it fail. Then we can use it to repair what ever is broken in the diffs.

gpt_engineer/benchmark/types.py
ATheorell: Has this been addressed? I think it happened because I ran few problems, only the first 5...
Or might it have been due to the catch clause that the list was empty?

azrv: I don't manage to reproduce it, but I added a check in the beginning of the method:

```
if not self.assertion_results:
            return 0.0
```

Can you still catch it?

gpt_engineer/benchmark/run.py
azrv: Done

azrv:
Solves https://github.com/gpt-engineer-org/gpt-engineer/issues/819

PR #1054, opened 2024-03-09
Image prompts

TheoMcCabe:
I was curious about the capabiities of adding vision to gptengineer. Ive successfully passed it a UX diagram as part of a prompt for changing a web app. It seems to have improved the quality of the outputted code (for this very specific case - reorganising the layout of a web page). One learning though is that the gpt-4-vision-preview is not handling the diff output as well and may not be as strong on software engineering in general? Regardless I'm pretty sure gpt-5 will have vision so this seems like a sensible long term direction to go in. 

The changes ive made: 
- prompts are now a Prompt object and not a string. They mirror the structure of a openAI / langchain message content. 
- prompts can still be provided with a prompt file in the repo - but there is a new format: a prompt folder containing a file called 'text' containing the text of the prompt and a folder called 'images' containing any images one wants to include in the prompt. Both methods still work as per my testing 
- Ai now has a boolean property 'vision'. Images are only included in the prompt if the AI is vision enabled. Currently this is true if the model name contains 'vision'. I imagine at some point langchain will provide a flag for this on different models which is more reliable.
- The tokenizer now contains custom code for converting images into token counts based on how openAI calculates images as tokens. This may not accurately convert over to other vision capable models so that's something to be aware of. tiktoken doesn't tokenizer images sadly. I may move this into a new open source project - forking tiktoken and adding vision capabilities to it. (TBC)
- added tests for image tokenizer

chores: 
- updated default model to gpt-4-0125-preview


AntonOsika:
This is very cool

I think we want to think through what the UX should be for this, and for it to not be confusing with "both file and folder is allowed" before merging..

(I think the UX of asking for prompt and then reusing it next time is _already_ quite confusing and should be addressed)


I think it's cool to merge it if we create a different application from cli?


Nitpicks:

- We should probably delete a lot of code and use `with get_openai_callback() as cb:` for token consumption, described [here](https://python.langchain.com/docs/modules/model_io/llms/token_usage_tracking)
- snake case please, for repoResult

TheoMcCabe:
@AntonOsika  thanks for the review. 

Re UX you mean the UX for the case when a prompt file hasn't been created? 

Yes good point - asking for one and then saving it is a bit odd. Do you want me to refactor so the prompt isnt saved as a file - and is just passed through for that run? 

Re file and folder - I find it analogous for the fact langchain allows a string or a big structured array so i dont mind accepting both. Most flexible approach would be to force using a folder but i dont want to break peoples existing workflows etc and thats quite a major change to make just for this! I suppose i would need to update a bunch of documentation too?

Im happy to just make a separate vision application for now if thats easiest

TheoMcCabe:
@ATheorell Any suggestions with these tests? I cant work out why their failing here but passing on main

TheoMcCabe:
> @ATheorell Any suggestions with these tests? I cant work out why their failing here but passing on main

no worries sorted it. That was painful - hopefully it was just a one off pain though because i broke caching by changing the way prompts are sent to the LLM

TheoMcCabe:
> * get_openai_callback()

Re this i wont have time to do that but i can create as an issue for someone else to pick up or i can come back to it when i have time. 

I might be able to find time to make it into a seperate CLI application just have a very busy week but ill see

ATheorell:
gonna find the time to test this soon. Sounds epic!

We will probably discuss this in the meeting on Thursday.

ATheorell:
Hi Theo,
we had a chat about the UX of this on the meeting on Thursday and I believe @ErikBjare had a great idea of how to go ahead with this. The cli should have two new optional arguments, one argument, lets call it "--prompt <path>", should allow the user to specify a custom prompt file (defaults to good ol' "prompt"). A second argument, "--image-repo <path>" should specify a directory with images (or directly point to an image file @TheoMcCabe ?).

TheoMcCabe:
> Hi Theo, we had a chat about the UX of this on the meeting on Thursday and I believe @ErikBjare had a great idea of how to go ahead with this. The cli should have two new optional arguments, one argument, lets call it "--prompt ", should allow the user to specify a custom prompt file (defaults to good ol' "prompt"). A second argument, "--image-repo " should specify a directory with images (or directly point to an image file @TheoMcCabe ?).

Yes this works! sounds good

TheoMcCabe:
Closing as @ATheorell asked me to re open on a branch on this repo so he can collaborate

PR #1055, opened 2024-03-09
Add support for claude 3

AntonOsika:


markojak:
how do we use this ?

TheoMcCabe:
> how do we use this ?

pass in the anthropic model you want to use as the second parameter when you call gpte 

e.g 

gpte . {model-name}

Issue #1050, opened 2024-03-07
The documentation build is failing


kristiankyvik:
The docs build started failing [about a week ago](https://readthedocs.org/projects/gpt-engineer/builds/). More concretely, it seems like they started breaking after [this PR got merged](https://github.com/gpt-engineer-org/gpt-engineer/pull/1039/files).

- I have tried to reproduce locally with no success
- I have looked at the files changed in PR #1039 but I am struggling to understand how those changes could be connected

Once this is fixed, we will build and serve the latest changes to the docs, that have already been merged.

kristiankyvik:
I am still struggling to reproduce this locally.

Based on [build logs,](https://readthedocs.org/projects/gpt-engineer/builds/) I see: 
- It does seems to installs the toml package
- But later raises the "ModuleNotFoundError: No module named 'toml'" error.  

My wild guess is that this could be related to the virtual environment setup, as in the venv we use to install the packages (including toml) is not the same one where we are executing the sphinx build command.

I do not have access to the instance running the documentation build, maybe @ErikBjare you can give me hand here?

kristiankyvik:
This has been fixed as per PR #1059

PR #1058, opened 2024-03-11
Updated the readthedocs build steps to account for Poetry 1.8 change regarding virtualenvs behaviour

kristiankyvik:
This PR should fix the issue #1050. 

In 1.7, poetry install added packages to the $READTHEDOCS_VIRTUALENV_PATH venv.
In 1.8, packages are added to the underlying Python env, not in the virtualenv.

As a consequence,  the packages added this way are not seen and the build fails. 

The solution is based on [this discussion](https://github.com/python-poetry/poetry/issues/9025#issuecomment-1963073941).

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -17,9 +17,8 @@ build:
   jobs:
     post_create_environment:
       - pip install poetry
-      - poetry config virtualenvs.create false
     post_install:
-      - poetry install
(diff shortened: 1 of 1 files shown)


ErikBjare:
This knowledge will come in handy for my other projects using poetry + RTD.

Thanks @kristiankyvik :heart:

PR #1068, opened 2024-03-18
build(deps): added tomlkit

ErikBjare:
Fixes broken tests for config due to missing dependency.

Why didn't it break CI though?


ATheorell:
Indeed strange that CI didn't catch this

ErikBjare:
@ATheorell I think since poetry depends on tomlkit, and poetry seems to be installed in the same venv as gpt-engineer (bad practice, but no biggie), it might have gotten `tomlkit` that way.

PR #1057, opened 2024-03-09
Add claude 3 / anthropic with cost calculation + convenience fixes. gpt-engineer is coauthor

AntonOsika:

PR #1070, opened 2024-03-18
ci: added mypy to precommit

ErikBjare:
While trying to get #1057 merged I noticed there were a bunch of type errors that were only caught in tests. I thought we had some kind of typechecking on the codebase, but apparently not.

I added mypy to the precommit config, but it throws 100+ type errors.

Here's the PR anyway, as a reminder we might want to get around to cleaning up the types.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,3 +25,10 @@ repos:
       - id: detect-private-key
       - id: end-of-file-fixer
       - id: trailing-whitespace
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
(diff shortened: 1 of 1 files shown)


ATheorell:
Good initiative! 

Wouldn't this be a nice project for automatic type fixing? Someone could set up a self-heal with gpt-engineer which for example fixes types file by file and runs until mypy passes.

ErikBjare:
@ATheorell Yes!

I think we should be able to make a general self-heal that supports any linter/checker.

Israel-Laguan (approved):

Issue #943, opened 2023-12-31
Instructions for running it with local models is lacking.


iswarpatel123:
## Policy and info
 - Maintainers will close issues that have been stale for 14 days if they contain relevant answers.
 - Adding the label "sweep" will automatically turn the issue into a coded pull request. Works best for mechanical tasks. More info/syntax at: https://docs.sweep.dev/

## Description
Instructions:

Running the Example[](https://gpt-engineer.readthedocs.io/en/latest/open_models.html#running-the-example)
Once the API is set up, you can find the host and the exposed TCP port by checking your Runpod dashboard.

Then, you can use the port and host to run the following example using WizardCoder-Python-34B hosted on Runpod:

  OPENAI_API_BASE=http://<host>:<port>/v1 python -m gpt_engineer.cli.main benchmark/pomodoro_timer --steps benchmark TheBloke_WizardCoder-Python-34B-V1.0-GPTQ
  
  What is this example? What does it do? Whats gpt_engineer.cli.main?
  
  How do i run the main command "gpte projects/my-new-project" after i have a local llm runing on localhost:8000?

## Suggestion
Please provide more step by step instructions.

viborc:
As a quick update for the community, we are actively working on this issue and experimenting with using several local models to see how well they can work with gpt-engineer. After that, based on our experiments, we will update the documentation with relevant info.

definitiontv:
I just got docker container working transparently using a dummy cloudflare hosted external address. On the server a  combination of 

ollama serve
litellm 

simulates openai

and in the .env file add OPENAI_API_BASE=https://ai.mydomain.com

I experimentally pointed ollama at mistral and code llama  - memory seems to produce code but no files are written so far.

Cloudflare tunnel is only because my implementation of docker does not recognize host.docker.internal and I didn't want to add a network to the docker compose.  But you could modify the docker compose to point to the hosts localhost:8000  if your setup supports that.

zigabrencic:
Hey. 

As discussed a proposal the local LLM support. Please provide feedback before I dive in.

**`ollama` and the rest can be added later following the same approach.**

Requirements:

- If possible don't add extra dependencies to `gpte`.
- Use `langchain` package for interacting with the LLM's.
- Adding minimal amount of code to `gpte` for the local LLM support.

# Support for `llama.cpp` compatible models

Custom LLM would be supplied to the `gpte` via `--open-llm` flag:

```bash
[...]

PR #1082, opened 2024-03-21
Support for Open LLMs

zigabrencic:
This PR resolves the [issue 943](https://github.com/gpt-engineer-org/gpt-engineer/issues/943) by adding an explanation on how to use `openLLMs`.

Overview of changes:

- minor change in `gpt_engineer/applications/cli/main.py` to work with open models.
- Rewritten `docs/open_models.md`
- Added `docs/examples` to help users test if their `LLM` setup works.

Feel free to let me know if anything is unclear or has to be modified.

Including the relevant reviewers: @ErikBjare @captivus @ATheorell @viborc


zigabrencic:
I managed to resolve the `tox/pytest` issues by re-naming the example scripts to `openai_api_interface.py` and `langchain_interface.py`

`pytest` thought that they were "tests" 😉

docs/open_models.md
```diff
@@ -1,24 +1,118 @@
...
+```bash
+python -m llama_cpp.server --model $model_path --n_batch 256 --n_gpu_layers 30
 ```
-  OPENAI_API_BASE=http://<host>:<port>/v1 python -m gpt_engineer.cli.main benchmark/pomodoro_timer --steps benchmark TheBloke_WizardCoder-Python-34B-V1.0-GPTQ
+
+Then in another terminal window set the following environment variables:
+
+```bash
+export OPENAI_API_BASE="http://localhost:8000/v1"
+export OPENAI_API_KEY="sk-xxx"
```
TheoMcCabe, line 100: Is this required? Can it not just null? Seems like an arbrirary magic string

gpt_engineer/applications/cli/main.py
```diff
@@ -76,8 +76,12 @@ def load_env_if_needed():
         load_dotenv()
     if os.getenv("OPENAI_API_KEY") is None:
         load_dotenv(dotenv_path=os.path.join(os.getcwd(), ".env"))
+
     openai.api_key = os.getenv("OPENAI_API_KEY")
 
+    if openai.api_key == "sk-xxx":
```
TheoMcCabe, line 82: this doesnt look clean to me. A boolean variable called isLocalModel or LocalModel or something would be better

docs/open_models.md
zigabrencic: According to this: https://llama-cpp-python.readthedocs.io/en/latest/server/#multimodal-models yes. 

Although I'm not sure why. Since we use `llama-cpp-python` for open model inference I decided to follow their documentation.

gpt_engineer/applications/cli/main.py
zigabrencic: I was thinking about this quite a lot and then decided that the above is cleanest. And requires the least amount of code changes in the code base. At the price of code readability  🤷‍♂️

If we introduce `LocalModel` variable or sth similar then we need to have a list of what clarifies as a local model. There are many options for open LLMs. Model list isn't as trivial as for the OpenAI case.

In principle we could require the user to set another env variable `local_model` and then decide on that.

Not sure. I'm open to suggestions here.

viborc:
@ErikBjare @captivus @ATheorell do you guys want to weigh-in on this PR, too?

gpt_engineer/applications/cli/main.py
viborc: @TheoMcCabe, do you have any follow-up comments on @zigabrencic's explanation?

ErikBjare (commented):
I haven't looked super closely, but looks good!

One thing that struck me when reading [Aider's blog](https://aider.chat/2024/03/08/claude-3.html) is that openrouter.ai has a OpenAI-compatible API which let's users dispatch requests to a wide range of models: from OpenAI and Anthropic models, to custom/open models via Together.ai and other providers.

This might be the fastest way to get started with open models, and doesn't require any special hardware. Worth mentioning in the docs imo!

zigabrencic:
As discussed in today's meeting I switched to `export LOCAL_MODEL=true`. 

And removed the rest except for this line: `elif os.getenv("LOCAL_MODEL"):`  which is needed to properly compute local API costs.

PR #1083, opened 2024-03-21
Retry diff logic

AntonOsika:
I think we should not merge this until @ATheorell thinks it is a good idea!

### DIFF
--- a/gpt_engineer/core/default/steps.py
+++ b/gpt_engineer/core/default/steps.py
@@ -312,10 +312,8 @@ def improve_fn(
 def _improve_loop(
     ai: AI, files_dict: FilesDict, memory: BaseMemory, messages: List
 ) -> FilesDict:
-    messages = ai.next(messages, step_name=curr_fn())
-
-    # check edit correctness
-    for _ in range(MAX_EDIT_REFINEMENT_STEPS):
+    for _ in range(MAX_EDIT_REFINEMENT_STEPS + 1):
+        messages = ai.next(messages, step_name=curr_fn())
         files_dict, errors = salvage_correct_hunks(messages, files_dict, memory)
         if not errors:
             break
--- a/gpt_engineer/core/default/steps.py
+++ b/gpt_engineer/core/default/steps.py
@@ -312,11 +312,11 @@ def improve_fn(
 def _improve_loop(
     ai: AI, files_dict: FilesDict, memory: BaseMemory, messages: List
 ) -> FilesDict:
-    for _ in range(MAX_EDIT_REFINEMENT_STEPS + 1):
-        messages = ai.next(messages, step_name=curr_fn())
-        files_dict, errors = salvage_correct_hunks(messages, files_dict, memory)
-        if not errors:
-            break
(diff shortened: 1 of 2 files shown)


gpt_engineer/core/default/steps.py
```diff
@@ -312,34 +312,31 @@ def improve_fn(
...
+    messages = ai.next(messages, step_name=curr_fn())
+    files_dict, errors = salvage_correct_hunks(messages, files_dict, memory)
+
+    retries = 0
+    while errors and retries < MAX_EDIT_REFINEMENT_STEPS:
+        messages.append(
+            HumanMessage(
+                content="Some previously produced diffs were not on the requested format, or the code part was not found in the code. Details:\n"
+                + "\n".join(errors)
+                + "\n Only rewrite the problematic diffs, making sure that the failing ones are now on the correct format and can be found in the code. Make sure to not repeat past mistakes. \n"
```
ErikBjare, line 324: I think this prompt could be better, here's the prompt we use for this in gptengineer.app:

```
async def fix_inapplicable_diffs(
    ai: AI,
    state_tracker: StateTracker,
    code: Dict[str, Union[str, bytes]],
    messages: list[BaseMessage],
    failed_diffs: list[Diff],
):
    """
    Given a list of diffs that could not be applied, ask the AI to fix them.
    """
    failed_diffs_context_string = (
        "Some of the diffs could not be applied. Here are the failed diffs:\n\n"
    )
    for failure in failed_diffs:
        failed_diffs_context_string += f"{failure.error_message}\n{failure.code_block.path}\n\n```{failure.code_block.language}\n{failure.format_git()}```\n\n"

    failed_diffs_context_string += """
Please write new diffs that can be applied to the code instead of the failed diffs.
If the content was found more than once, include one surrounding context line before and after the content.
If the content was not found, please provide the correct content.
    """.strip()

    messages.append(ai.fuser(failed_diffs_context_string))
    return await generate_and_apply_diffs(
        ai, state_tracker, code, messages, allow_fixing_inapplicable_diffs=False
    )
```

https://github.com/lovablelabs/lovable/blob/903a2d94bf9b358053e78e8b1b58c721aa5bac8c/api/core/agents/gpt_engineer/improve_steps.py#L61-L86

It relies on informative errors from the diff parser. The main two cases are no matches and multiple matches (non-unique).

ATheorell: Found more than once should not be a problem here. This rejection explicitly means that it does not conform to any diff possible diff in the original file

ErikBjare: @ATheorell Why is found more than once not a problem? Is it not checked for?

ATheorell: Unified git diffs contain line numbers in the header. Line numbers only occur one time :)

ATheorell: The line number in the diff tends to be correctly given by the LLM, since it is read directly from the source file. The remaining three numbers in the header are not trustworthy, but they can be computed and corrected from the content of the diff

PR #1094, opened 2024-04-01
Formate uploading files with black linting tool

similato87:
## Overview

This PR aims to address Issue #1023 by integrating code formatting and linting tools into our development workflow. The primary goal is to enhance efficiency and improve the quality of the suggestions provided by LLMs by pre-screening code for errors. This preemptive check reduces the computational demand on LLMs, allowing them to focus on the complexity of tasks rather than basic syntax corrections.

## Implementation Details

To minimize the impact on user experience, we plan to implement an automatic linting feature that will be enabled by default during the code processing phase. Users will be informed that their codebase has undergone linting, ensuring transparency. Additionally, we will provide an option for users who prefer not to use the linting tools, respecting diverse development preferences.

## Scope and Targets

The initial rollout will focus on supporting Python files. This decision is made to strike a balance between managing dependencies and meeting the feature's requirements efficiently. By starting with these languages, we aim to cover a significant portion of the development activities within our community while keeping the implementation manageable.


DannyFranklin (approved):

ATheorell:
I have a extra request :)

I think it would make sense to always run linting by default inside FileStore:push and also run it by default inside FileStore:pull, but in this case, make an optional argument to disable linting

similato87:
> I have a extra request :)
> 
> I think it would make sense to always run linting by default inside FileStore:push and also run it by default inside FileStore:pull, but in this case, make an optional argument to disable linting

I support this suggestion as it aligns with best practices for software development. I'm planning to refactor this section in my upcoming PR. Would it be advisable to introduce a return flag at this point in the code? [main.pyL456](https://github.com/gpt-engineer-org/gpt-engineer/blob/673a72afd05b82baeaa79563537d8dc275366e38/gpt_engineer/applications/cli/main.py#L456C1-L456C75). I'm considering modifying the function to return `files_dict_before` and `isLinting` like so: `files_dict_before, isLinting = FileSelector(project_path).ask_for_files()`. Additionally, should I initialize a FileStore in the main function to handle both pull and push operations, utilizing the `isLinting` flag to toggle the linting status?

ATheorell:
> should I initialize a FileStore in the main function to handle both pull and push operations, utilizing the `isLinting` flag to toggle the linting status?

Yes, that sounds great! Actually, such a FileStore object is already initialized in the main at line 453 and could be used for all io-operations. Inside the FileSelector, it would then make sense to use the isLinting variable to toggle whether the FileStore object should lint or not for the case of reading files for improve.

PR #1096, opened 2024-04-01
fix: incorrect prompt type breaking gptme benchmarks

Styren:
The gptme benchmarking suite uses strings for prompts which causes an error because `agent#improve` assumes it's a `Prompt` object.


viborc:
Looks good to me! @ErikBjare, can I ask you to take a look and see if this is fine?

ErikBjare (approved):
LGTM 👍

PR #1098, opened 2024-04-02
Ollama API endpoint parallel agent Visualization

6rz6:
Added the infrastructure including the v1/completions Ollama Endpoint (Open AI compatible, just replace the APIs endpoint) Multi threaded loading of the Models from a local or remote server. (In progress) Parallel Visualization of 4 active agents including display of resource consumption per each Ollama model in real time Use of Quantized, highly focused trained mixture of models to achieve many correct fast interactions thus improving the final desired result - i call it ("Decentralized Mixtral 8x7b") <- which is by far my favorite model - by mistral AI

### DIFF
--- a/Ollama_Models_Arena.py
+++ b/Ollama_Models_Arena.py
@@ -1,0 +1,128 @@
+import tkinter as tk 
+from tkinter import ttk
+import threading
+import time
+import requests
+import subprocess
+
+class LLMTesterApp:
+
+  def get_ollama_models(self):
+    result = subprocess.run(['ollama', 'list'], capture_output=True)
+    lines = result.stdout.decode().splitlines()  
+
+    models = []
+    for line in lines:
+      parts = line.split("\t")
+      model_name = parts[0]
+      models.append(model_name)
+
+    return models
+
+  def __init__(self, master):
+    self.master = master
+    self.master.title("LLM Response Time Tester")
+
+    self.models = self.get_ollama_models()
+    self.setup_ui()
+
+  def setup_ui(self):
+    
+    self.output_frames = []
+    self.output_texts = []
+    self.model_dropdowns = []
+    self.timers = []
+
+    for i in range(4):
+      frame = ttk.Frame(self.master)
+      frame.grid(row=1, column=i, padx=10, pady=10)
+
+      model_var = tk.StringVar()
+      dropdown = ttk.Combobox(frame, textvariable=model_var, values=self.models)
+      dropdown.grid(row=0, column=0, padx=5, pady=5)
+      self.model_dropdowns.append(model_var)
+
+      output_text = tk.Text(frame, width=30, height=15)
+      output_text.grid(row=1, column=0, padx=5, pady=5)
+      self.output_texts.append(output_text)
+
+      timer_label = ttk.Label(frame, text="00:00")
+      timer_label.grid(row=2, column=0, padx=5, pady=5)
+      self.timers.append(timer_label)
+
+      self.output_frames.append(frame)
+
+    input_frame = ttk.Frame(self.master)
+    input_frame.grid(row=2, columnspan=4, padx=10, pady=10)
+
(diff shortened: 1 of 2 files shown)


captivus:
Hello!  Please clarify the features you propose to add via this PR, along with their specific benefits to GPT-Engineer.  Also, please ensure that all of the checks are successful.

6rz6:
Hi, I'm adding optimized parallel Ollama model loading, with an optional UI to assess the load the the models on the system.
The goal is to improve the performance of multiply agent working in parallel each with the ability to use a task specific fine tuned quantized LLM.
This will allow us to assess the minimal system requirements to run in the future on 'smaller' devices and get results in an adequate time frame, while recommending the user based on his HW what are the Ollama models which will perform optimally on his device.
Feel free to brainstorm with me, and any feedback would be highly welcome.

6rz6:
running the exact same command on my localmachine and it passes all green:
image

I dont get it, need to understand what is being checked differently with github pre-commit

6rz6:
I still get no hook id, will investigate more
image

6rz6:
HI how are you?
I believe i found the reason why check 0.3 failed, it the alias of my py Vs the python ver of the alias of your py .
i run 3.10.13 when running any python and the one on your server was set to 3.8.
as you wrote in the branches the support for v.23 was the last for python 3.8 and i was running all this rime on 3.10.13 and failing a test which all it does is basically to check the support of the file in python, in this case it was tes3 and that drove me insane all last night.

i can run it locally and i can also call the module locallyand in both cases since it m interpreter everything woek fine. just we when tried to run the python test. we were border line because i assume you just recently upgraded to py3.10-12 and probably it slipped of your mind.

I managed to run it today both from your server and my local both went smooth. all i had to do to solve the issue is to change *inmy session ofc not in your bashrc, the aliass of your py to 3.10.11.

i tried to do more tests today but the module is already in main which i cannot acesss and also i lost access to the branch you opened for me... :)

talk to me later when available, we can sync a bit

6rz6 (commented):
** #The code works well, tested it today on multiple machines**

README.md
```diff
@@ -1,4 +1,4 @@
-# GPT-Engineer
+# Ollama-Multi-threaded-UI-Integration for GPT-Engineer
```
6rz6, line 1: changed from @6rz6 [Ollama_Models_Arena.py](https://github.com/gpt-engineer-org/gpt-engineer/pull/1098/files#diff-28d7c8481b3d39311b7059628a1d9020b6d8a6e95e87e0815ac24976d4c6cc31) 

Integrated on v.0.28->0.29

6rz6:
ok, passed and integrated

[Tox pytest all python versions](https://github.com/gpt-engineer-org/gpt-engineer/actions/workflows/ci.yaml)

Ollama_Models_Arena.py
6rz6, line 1: @6rz6  [6rz6:6rz6-Ollama-Multi-threaded-UI-Integration](https://github.com/gpt-engineer-org/gpt-engineer/pull/1098/files#diff-28d7c8481b3d39311b7059628a1d9020b6d8a6e95e87e0815ac24976d4c6cc31)
> 
> [Ollama_Models_Arena.py](https://github.com/gpt-engineer-org/gpt-engineer/pull/1098/files#diff-28d7c8481b3d39311b7059628a1d9020b6d8a6e95e87e0815ac24976d4c6cc31)
> 
> Integrated on v.0.28->0.29

README.md
TheoMcCabe: Your changing the title of the readme?

TheoMcCabe:
I dont understand the point of this code to be honest.  What does it have to do with GPT Engineer?

ErikBjare:
I don't think this makes sense to merge either. We're trying to keep the repo as lean as possible, and this has very little connection with gpt-engineer itself.

If you want to make this a tool for others to use, consider making your own repo :)

Issue #1076, opened 2024-03-19
Disentangle tests by removing cachingAI


ATheorell:
Some time back I made the unfortunate decision of implementing a "CachingAI" class that allows us to run tests here on github without an openai api key. It turns out that, using CachingAI in unit tests, instead of simpler mocks makes the tests highly tangled -- changing one thing often leads  to a plethora failed tests and its hard to trace why they fail -- the opposite of the intention of unit tests. 

To remedy this, we should remove the CachingAI class completely and replace its usage in the tests with mocks, EXCEPT for the top level tests, which should be evaluated live with openai_key (yes we can use that on github).

The specific tests that should be _deep_ are the top level tests in test_main.py, specifically:

    def test_default_settings_generate_project(self, tmp_path, monkeypatch):

    def test_improve_existing_project(self, tmp_path, monkeypatch):

    def test_lite_mode_generate_project(self, tmp_path, monkeypatch):

    def test_clarify_mode_generate_project(self, tmp_path, monkeypatch):

    def test_self_heal_mode_generate_project(self, tmp_path, monkeypatch):

and in test_install.py:

    def test_installed_main_execution

ErikBjare:
We should consider using pytest markers to annotate the tests that rely on inference (or simply use a conditional skip if there is no OPENAI_API_KEY in the env), so that they can be toggled.

ATheorell:
> We should consider using pytest markers to annotate the tests that rely on inference (or simply use a conditional skip if there is no OPENAI_API_KEY in the env), so that they can be toggled.

Makes sense

PR #1104, opened 2024-04-04
Disentangle tests #1076

ATheorell:
As stated in #1076 the tests are way too tangled. In particular, usage of cached chatgpt responses in the tests have led to that most small changes have lead to multiple failing tests. This has been particularly challenging for new contributors.

In this PR:
- No more caching in tests.
- test_main.py no longer tests any workflow execution, only setup.
- A single test in test_install.py runs the installed gpte executable WITH api key, no other test does. This test is marked with "requires_key" after suggestion by @ErikBjare. No other tests require api key.
- Testing of running full improve workflows from the main currently missing, should be satisfied with specific unit tests of file_selector.py


ATheorell:
Tox fails with exception that it cannot find valid OPENAI_API_KEY, though I've tried adding it in the ci.yaml

name: Run tox
env:
  OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
run: tox
      
     
There is a secret called OPENAI_API_KEY

Do you have an idea what could be going wrong here @ErikBjare ? I saw that you have a rather advanced setup including a key in https://github.com/ErikBjare/gptme/blob/master/.github/workflows/bot.yml

ErikBjare:
@ATheorell I misunderstood, I'm not sure why it isn't working as it currently looks.

I figured it must be a quirk of tox, and I found this: https://tox.wiki/en/latest/config.html#passenv

So we need to set `passenv=OPENAI_API_KEY` (or whatever the syntax is), for tox to pass along the environment variable.

ErikBjare:
@ATheorell Took the liberty, we'll see if it works...

PR #1127, opened 2024-04-24
Update .readthedocs.yaml

Harry1402:


viborc:
Thanks for this, but this is just a cosmetic PR and based on our policy, it's being rejected.

PR #1137, opened 2024-04-28
code to test whether tox is culprit for openai_key fail on github

ATheorell:


ATheorell:
Not intended to be merged at all

PR #1146, opened 2024-05-11
extract linting process from file_selector

similato87:
As discussed in #1094, extracted the linting from the file selector process to enhance modularity and align with our software design patterns. Please review the updated implementation.


ATheorell:
Thanks!

Like we discussed, it would be more intuitive to set the linting flag in a general config. Merging for now!

Issue #1136, opened 2024-04-28
Docker build is broken


t6tv8e:
## Policy and info
 - Maintainers will close issues that have been stale for 14 days if they contain relevant answers.
 - Adding the label "sweep" will automatically turn the issue into a coded pull request. Works best for mechanical tasks. More info/syntax at: https://docs.sweep.dev/

## Expected Behavior

Please describe the behavior you are expecting.

After running `docker build --rm -t gpt-engineer -f docker/Dockerfile .` I should have a runnable container.

## Current Behavior

Unable to build the container due to PyArrow and CMake issues.

### Failure Logs
```
  creating /tmp/pip-install-673hp2i9/pyarrow_d35c22e7f3e947faaa29331bd28d4241/build/temp.linux-x86_64-cpython-311
43.03       -- Running cmake for PyArrow
43.03       cmake -DCMAKE_INSTALL_PREFIX=/tmp/pip-install-673hp2i9/pyarrow_d35c22e7f3e947faaa29331bd28d4241/build/lib.linux-x86_64-cpython-311/pyarrow -DPYTHON_EXECUTABLE=/usr/local/bin/python -DPython3_EXECUTABLE=/usr/local/bin/python -DPYARROW_CXXFLAGS= -DPYARROW_BUILD_CUDA=off -DPYARROW_BUILD_SUBSTRAIT=off -DPYARROW_BUILD_FLIGHT=off -DPYARROW_BUILD_GANDIVA=off -DPYARROW_BUILD_ACERO=off -DPYARROW_BUILD_DATASET=off -DPYARROW_BUILD_ORC=off -DPYARROW_BUILD_PARQUET=off -DPYARROW_BUILD_PARQUET_ENCRYPTION=off -DPYARROW_BUILD_AZURE=off -DPYARROW_BUILD_GCS=off -DPYARROW_BUILD_S3=off -DPYARROW_BUILD_HDFS=off -DPYARROW_BUNDLE_ARROW_CPP=off -DPYARROW_BUNDLE_CYTHON_CPP=off -DPYARROW_GENERATE_COVERAGE=off -DCMAKE_BUILD_TYPE=release /tmp/pip-install-673hp2i9/pyarrow_d35c22e7f3e947faaa29331bd28d4241
43.03       error: command 'cmake' failed: No such file or directory
43.03       [end of output]
43.03
43.03   note: This error originates from a subprocess, and is likely not a problem with pip.
[...]

PR #1159, opened 2024-05-23
1136 docker build is broken

captivus:
This solves issue #1136 by migrating the Docker image to use `python:3.11-slim`, as building `pyarrow` is broken in Alpine Linux.

I'm not aware of the specific reason that `apk` was chosen as the previous image, but this PR replaces it with a working container that successfully executes `gpt-engineer` CLI.

Added `langchain-community` to Poetry to resolve the application (container) crashing upon startup.

We should discuss container optimization, if that remains a concern.


zigabrencic:
Looks good to me since all the points from [Issue 1136 discussion]( https://github.com/gpt-engineer-org/gpt-engineer/issues/1136#issuecomment-2118779798) are considered.

I tested it on MacOS and it works.

docker/Dockerfile
```diff
@@ -10,7 +15,9 @@ COPY . .
 RUN pip install --no-cache-dir -e .
 
 # Stage 2: Final stage
-FROM python:3.11-alpine
+FROM python:3.11-slim
+
+RUN apt-get update && apt-get install -y --no-install-recommends git
```
k1lgor, line 20: This git installation is not followed by the best practice of using a multi-stage build in Docker. The purpose of multi-stage builds is to reduce image size, improve security, simplify the build process, and enhance maintainability. I have posted a proper Dockerfile in my last comment on #1159

k1lgor:
> This solves issue #1136 by migrating the Docker image to use `python:3.11-slim`, as building `pyarrow` is broken in Alpine Linux.
> 
> I'm not aware of the specific reason that `apk` was chosen as the previous image, but this PR replaces it with a working container that successfully executes `gpt-engineer` CLI.
> 
> Added `langchain-community` to Poetry to resolve the application (container) crashing upon startup.
> 
> We should discuss container optimization, if that remains a concern.

To be clarify, `pyarrow` is supported on Alpine, it lacks a **_wheel_**. In other words, you need to build it from source.

docker/Dockerfile
captivus: I don't see the comment you're linking to, @k1lgor, as you're referencing this same PR.  I'm not a Docker expert, so I'd welcome guidance if you have it.

captivus:
> > This solves issue #1136 by migrating the Docker image to use `python:3.11-slim`, as building `pyarrow` is broken in Alpine Linux.
> > I'm not aware of the specific reason that `apk` was chosen as the previous image, but this PR replaces it with a working container that successfully executes `gpt-engineer` CLI.
> > Added `langchain-community` to Poetry to resolve the application (container) crashing upon startup.
> > We should discuss container optimization, if that remains a concern.
> 
> To be clarify, `pyarrow` is supported on Alpine, it lacks a **_wheel_**. In other words, you need to build it from source.

Yes, I had written a previous version of the Dockerfile which successfully compiled `pyarrow` from source but it would not install as a Python package.  After some time mucking about trying to resolve that, I decided to change the distro for to enable a prompt resolution of this issue that has been open for longer than we'd like.

docker/Dockerfile
k1lgor: Here is the Docker file I have posted:
```bash
# Stage 1: Builder stage
FROM python:3.11-slim AS builder

RUN apt-get update && apt-get install -y --no-install-recommends \
  tk \
  tcl \
  curl \
  git \
  && rm -rf /var/lib/apt/lists/*

WORKDIR /app

COPY . .

RUN pip install --no-cache-dir -e .

# Stage 2: Final stage
FROM python:3.11-slim

WORKDIR /app

COPY --from=builder /usr/local/lib/python3.11/site-packages /usr/local/lib/python3.11/site-packages
COPY --from=builder /usr/local/bin /usr/local/bin
COPY --from=builder /usr/bin /usr/bin
COPY --from=builder /app .

COPY docker/entrypoint.sh .

ENTRYPOINT ["sh", "/app/entrypoint.sh"]
```
The difference is that the git is installed at the building stage instead of the final, and since git and the other packages are installed via the OS package manager, the binaries are located in the `/usr/bin/` folder and just copy that folder in the final stage. This allows `gpte` to use `git`.

captivus: Ahhh, I see.  I've just built the container using the file you've provided and ... **it works!**

I wasn't trying to front run your solution with this PR.  I'll kindly offer a bit of reciprocal best practice advice for submitting solutions to issues.  ;)  The best practice is to submit a pull request containing your solution and then solicit feedback on the PR directly, rather than pasting code into an issues discussion.

We'd welcome this PR from you and, given that your solution predates mine, I'm happy to close this PR (and will do so)!  Will you please submit a pull request today?  We discussed this issue during our dev meeting yesterday and were keen to resolve it as it's been broken for longer than we'd like.

captivus:
Closing, having requested a PR from @k1lgor.

docker/README.md
```diff
@@ -1,24 +1,54 @@
...
+### Using Docker Compose
+
+1. **Clone the Repository** (if not already done)
+   ```
+   git clone https://github.com/gpt-engineer-org/gpt-engineer.git
+   cd gpt-engineer
+   ```
+
+2. **Build and Run using Docker Compose**
+   ```
```
k1lgor, line 40: Here are some alternatives using `docker compose`:
Since there is only one `docker-compose.yml` file, you could run it without the `-f` option.
- `docker compose up -d --build` - Build images before starting containers, when they are built, run the containers in the background
- `docker compose up -d` - Only start the containers in the background
- `docker compose down` - Stops containers and removes containers, and networks
- `docker compose down -v` - Removes the volumes as well

docker/Dockerfile
captivus: One more thing -- please update the README for the Docker container in a similar manner as I have.  Thanks!

k1lgor: Yeah, sure. I will prep the PR as soon as possible.

docker/README.md
captivus: Love it!  Please incorporate in the updated README in your forthcoming PR.

captivus:
After doing some digging, I see, now, that you are the great community member who originally rewrote the Dockerfile for this [a couple of months back](https://github.com/gpt-engineer-org/gpt-engineer/commit/d5a73ab3bdad7543f81b910ad23131d3a6877ba6) to make it lean and mean!  Please do build upon your prior good work.

Also, we'd love to have you contribute more broadly to the project, if you're keen.  We meet every Thursday at 19:00 CEST on [Discord](https://discord.gg/8tcDQ89Ej2) to collaborate.  Will you join us?

k1lgor:
@captivus Sure, I'd love to contribute beyond my full-time job. Although I may not be very active on Discord, I'll be happy to join the Thursday meetings whenever possible.

PR #1161, opened 2024-05-29
Hotfix/docker image

k1lgor:
# What's changed
- `docker-compose.yml` - version is deprecated
- `docker/Dockerfile` - image changed to slim base (Ubuntu)
- `docker/README.md` - update with Docker Compose instructions
- `pyproject.toml` - add `langchain-community` python package

This would fix #1136 and maybe #1160

P.S. Sorry for the delayed PR. Everything discussed in #1159 and #1136 implemented
@viborc @captivus @zigabrencic

PR #1162, opened 2024-05-31
Making langchain community explicit dependency

ATheorell:
langchain community is not a dependency of the newer versions of langchain, however, it is a dependency of gpt-engineer.  This means that, if we update the lock.toml file, we may get a langchain version that doesn't include community, which breaks gpt-engineer. This PR fixes this, by adding langchain-community as an explicit dependency


k1lgor:
I have already added this package in #1161

PR #1175, opened 2024-06-21
Update README.md

yimothysu:
Fix typos/grammar


viborc (approved):
While this would generally be considered something we call a cosmetic or a pedantic change, I'll approve this as @yimothysu and I had a call on Friday and discussed this in a broader context.

PR #1182, opened 2024-07-07
fixing tox Python 3.11 and 3.12 issues

captivus:
Python 3.11 target doesn't run in `tox`. Fixing this.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ on:
     branches: [ main ]
 
 concurrency:
-  group: ${{github.workflow}} - ${{github.ref}}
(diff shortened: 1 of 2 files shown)


captivus:
The error seems to be due to `codecov`. Disabling it results in successful run of Python 3.11 target. 

The remaining Python 3.12 test errors were due to this Pydantic issue which requires downgrading to Python 3.12.3 to resolve until Pydantic upstream fixes this for > Python 3.12.3.  Pinned our tox test installed version of Python 3.12 to 3.12.3.

captivus:
@ATheorell @viborc we need to sort out the `codecov` rate limit issue. I don't have access to the Codecov account to resolve.  When resolved, we can reenable `codecov`.

